PR #73, opened 2014-11-11
commented /utilities/option.cc

bakerte:
First comments for ITensor in /utilities/option.cc.  Geared towards the broadeest possible audience.


emstoudenmire:
I'm going to close this since (1) we've made a lot of changes to this file since this pull request and (2) I think it's better to have documentation on the website rather than too many comments in the code itself.

PR #77, opened 2015-02-27
Fixed paths in CMake files; added option to build static libs

poboiko:
Instead of installing 3 static libraries (libraries with quite common name like libmatrix/libutility inside /usr/lib might be confusing!), build single shared library libitensor.so and add a CMake property to build a static library libitensor_static.a.
This also fixes some paths for headers (e.g it was looking for non-existent tinyformat.h inside itensor/ directory while it was moved to utilities/)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required (VERSION 2.8)
 project (ITensor)
 
+option(BUILD_STATIC "Build static libraries" OFF)
+
 # Disable build in source (not to overlap with existing Makefile)
 set(CMAKE_DISABLE_SOURCE_CHANGES ON)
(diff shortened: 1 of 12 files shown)


emstoudenmire:
Hi thanks for these good suggestions. I'm working on a new version and will take all of these into account. I'm not sure about doing the single static lib change to the v1.x.x versions since it would be a (mildly) breaking change. But it's a nice idea for the 2.0 version which should hopefully be out sometime this summer.

emstoudenmire:
Hi I hope you don't mind but I'm going to close this. I've been wondering whether cmake is really necessary for us to support as our number of dependencies has dwindled to just blas and lapack and a C++ compiler. I think at this point it's actually easier for users to find the folder where their lapack is installed by themselves than going through all the steps of installing a using cmake, which often fails to find lapack anyway! But thanks for posting this update.

PR #78, opened 2015-05-11
matrix/lapack_wrap.h: replaced variable-length arrays with std::vector

kyungminlee:
Variable-length arrays are currently not C++ standard.

### DIFF
--- a/matrix/lapack_wrap.h
+++ b/matrix/lapack_wrap.h
@@ -8,6 +8,7 @@
 //
 // Headers and typedefs
 //
+#include <vector>
 
 #ifdef PLATFORM_lapack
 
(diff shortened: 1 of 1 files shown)


emstoudenmire:
Hi Kyungmin,
Great catch thanks. I just posted a new release of ITensor (v1.1.0) today which includes this fix (and I credited you on the changelog). I just merged your pull request but made the code slightly different way by making the vectors static and calling resize on them each time the function is run to avoid allocations when the workspaces keep the same size or are resized downward.

emstoudenmire:
Hm so I merged it manually into the master branch but it's not showing it closed, I think because you did the changes on the develop branch. Anyway I will mark it as closed now.

PR #83, opened 2015-05-26
Changes for MSVC compatibility

kyungminlee:
I confirmed , with these changes ITensor can be compiled with cmake + MSVC + MKL using the following command

```
cmake.exe E:\dev\src\ITensor -G "Visual Studio 12 2013 Win64" -DSample=OFF -DTutorial=OFF -DTesting=OFF -DBLA_VENDOR=Intel
```


emstoudenmire:
I'd really appreciate if you could check that this patch still works as I applied it to the latest version of ITensor, thanks.

PR #88, opened 2015-10-21
New AutoMPO implementation

akeselman:
Works for general N-site terms and optimizes the dimension of the MPO in case of long range interactions.


emstoudenmire:
Code was merged via a different set of patches. Great work by Anna!

PR #90, opened 2016-01-16
change static const in namespace scope to const

mingpu:
Static in namespace scope is superfluous. So I change "static const" to "const" in 
* global.h
* mpo.h
* real.h


emstoudenmire:
Also applied this change to the version 2.0 branch

PR #93, opened 2016-04-02
Update myappname.cc

mingruyang:
Change a line
`T = randomize(T);`
to
`randomize(T);`


emstoudenmire:
Great catch - thanks Mingru

PR #95, opened 2016-04-08
bring toIQMPO, toIQ back to work

xichuang:
Update toIQMPO(), toIQ(), convertToIQ() to itensor version 2 .
Those codes are based on the original codes in v1, and tested with the programs in fold samples by substituting
`auto H = IQMPO(ampo)`

with

`
auto A = MPO(ampo);
auto H = A.toIQMPO();
`
which give the same results.

### DIFF
--- a/itensor/mps/mpo.h
+++ b/itensor/mps/mpo.h
@@ -68,8 +68,8 @@ class MPOt : private MPSt<Tensor>
     MPOt<ITensor>
     toMPO() const;
 
-    //MPOt<IQTensor>
-    //toIQMPO() const;
+    MPOt<IQTensor>
+    toIQMPO() const;
 
     //MPOt: index methods --------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


emstoudenmire:
Hi, thanks for contributing this and sorry it's been so long. Part of the issue is that while this functionality is nice to have, I wasn't sure about maintaining it myself and supporting it as a core part of ITensor. The toIQMPO feature was always a tricky because it relies on a lot of heuristics such as when to treat a very small number as exactly zero, etc. So I had quietly stopped supporting it in the new version. But people may still benefit from it in certain cases.

So we came up with a new way for people to contribute complex, somewhat experimental codes like these. We now have a "Codes" page on the ITensor site: http://itensor.org/codes
If you post your toIQMPO code in its own repo on Github, we will be glad to link to it and list you as the maintainer. Then people can use it for their own projects, we can all observe if the code is working well, and then possibly decide to include it in ITensor later. In some sense, it is like a set of "plugins" for ITensor, similar to Julia or python packages. 

Please let me know what you think about posting your toIQMPO implementation as a separate code this way - thanks.

PR #100, opened 2016-04-15
Use size_t instead of unsigned long.

kyungminlee:
size of unsigned long does not match size of itensor::Index for LLP64 systems like Windows, resulting in static_cast errors. Use size_t instead.


emstoudenmire:
I see, so on Windows long and int are the same? Good catch & thanks for the fix.

PR #101, opened 2016-04-16
Replaced macro _MSC_VER with _WIN32 to support MinGW.

kyungminlee:


emstoudenmire:
Great, thank you for these Windows related fixes. I should work on getting access to a Windows computer so I can test ITensor on one myself in the future.

kyungminlee:
No problem. Unfortunately itensor fails to build with Microsoft Visual C++, even with the most recent version released yesterday. I think it is related to bugs in the compiler itself related to C++11 support. I've checked that with msys + mingw64 + OpenBLAS, itensor builds successfully and all the codes in "sample" produce the same results as in Linux or MacOS.

PR #104, opened 2016-04-27
Fixed error.h to #include <cstdlib>

kyungminlee:
abort() is declared in cstdlib.


emstoudenmire:
Always good to know about these since every platform is different

PR #105, opened 2016-06-19
Restore cmake

aeantipov:
A revised version of CMake for ITensor. Now has only one file for better maintenance.


aeantipov:
Miles, don't forget about this )

PR #106, opened 2016-06-29
Minor change of the file sweeps.h

mingpu:
Minor change of the file sweeps.h. Now if the number of row of sweep parameters in the input file is less than nsweep, the last row will repeat.

### DIFF
--- a/itensor/mps/sweeps.h
+++ b/itensor/mps/sweeps.h
@@ -360,10 +360,24 @@ tableInit(InputGroup& table)
     noise_ = std::vector<Real>(nsweep_+1);
 
     table.SkipLine(); //SkipLine so we can have a table key
+    int n_last = 1;
     for(int i = 1; i <= nsweep_; i++)
         {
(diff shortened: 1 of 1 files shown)


emstoudenmire:
Hi Mingpu, this is a nice idea. One question is, this seems to rely on a certain behavior of the line
    table.file() >> maxm_[i] >> ...
namely that it will set maxm_[i] to zero (or leave it as zero) if there is no more input. Is this relying on a particular guaranteed behavior of the >> operator for input streams? What if someone leaves a space between the last line in the table and the closing brace (closing } )? I just want to better understand if the condition we are relying on here is robust. Thanks!

mingpu:
Hi Miles, thanks. I think ifstream ignores white spaces. So if there are white spaces at the end of the file, the value of maxm_[i] will be the default value when it is declared, which is 0 for std::vector\<int>.

emstoudenmire:
Thanks Mingpu - I hope it doesn't lead to any surprising behavior but it is definitely an improvement to the current behavior!

PR #107, opened 2016-08-07
Fixed parameter types for OpenBLAS version of cblas_zdotc_sub in zdot…

kyungminlee:
…c_wrapper.


emstoudenmire:
Thanks for keeping an eye on these things! The OpenBLAS support was experimental so it may have a few glitches unfortunately.

PR #108, opened 2016-08-25
Lattice generation

ybarlev:
Corrected lattice bonds generation for 1D chains and ladders.
Previously extra bonds were added for periodic boundary conditions and Ny<=2


emstoudenmire:
Nice improvement, thanks

PR #109, opened 2016-09-20
Fixed typo in detail::checkArrows

lfrahm:
Found a typo in the detail::checkArrow function and fixed it. Now it prints the correct IndexSet to the console when IQIndices do not have matching arrows. Before, it was giving the same IndexSet twice.

### DIFF
--- a/itensor/itensor_operators.cc
+++ b/itensor/itensor_operators.cc
@@ -64,7 +64,7 @@ checkArrows(IQIndexSet const& is1,
                 println("----------------------------------------");
                 println("IQIndexSet 1 = \n",is1);
                 println("----------------------------------------");
-                println("IQIndexSet 2 = \n",is1);
+                println("IQIndexSet 2 = \n",is2);
                 println("----------------------------------------");
                 printfln("Mismatched IQIndex %s",I1);
                 Error("Mismatched IQIndex arrows");
--- a/itensor/mps/sites/hubbard.h
+++ b/itensor/mps/sites/hubbard.h
@@ -176,6 +176,14 @@ class HubbardSite
             Op.set(UD,UDP,+1);
             }
         else
+        if(opname == "1")
+            {
+            Op.set(Em,EmP,+1); 
+            Op.set(Up,UpP,+1);
+            Op.set(Dn,DnP,+1);
+            Op.set(UD,UDP,+1);
+            }
+        else
         if(opname == "Sz")
             {
             Op.set(Up,UpP,+0.5); 
--- a/itensor/mps/sites/hubbard.h
+++ b/itensor/mps/sites/hubbard.h
@@ -176,14 +176,6 @@ class HubbardSite
             Op.set(UD,UDP,+1);
             }
         else
-        if(opname == "1")
-            {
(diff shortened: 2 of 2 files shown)


emstoudenmire:
Hi, thanks for the fix to the error message. Regarding adding the identity operator, the SiteSet class already makes the identity operator (name is "Id") for all of the other site set types (see the end of this documentation page: http://itensor.org/docs.cgi?page=classes/siteset ). So if you remove that change I will merge this pull request.

lfrahm:
Hi, thanks for the note. I didn't mean to include the identity operator commit into this pull request, this happened by accident. Sorry. I reverted it and I think it is now good to go. Thanks!

Another question I have: I have implemented the == operator for Tensors. It only works for the QDense data structure so far, but if you are interested I could contribute that operator working for all types of tensors. Just let me know if you want me to build that.

emstoudenmire:
Ok merged, thanks. I didn't define operator== for tensors because I don't know of a use case for it. What use case did you find for it and how is it defined? Does it just compare the indices or also the numerical elements of tensors? If it compares the elements, what floating point error tolerance does it use when comparing floating point numbers?

lfrahm:
Great, thanks for merging! I was missing the == operator while working on a quantum chemistry MPO. It was mostly debugging stuff, but it turned out to be quite useful. You can find it in the tnsrcmpr branch of my fork: https://github.com/lfrahm/ITensor/tree/tnsrcmpr. It compares the values as well. (Since I have been using pre-defined values so far it compares the values with the == operator, of course that has to change when getting into the numerics)

I tried to implement it similar to the other operators but please keep in mind I'm still understanding the design and the data structures. So it might not be the best implementation right now, but I'm willing to fix that and make it generic if you are want me to.

emstoudenmire:
I agree it would be a useful thing to have. However, I don't think it should be called operator== because it could confuse many people. On a more technical level, operator== can't take any additional parameters. A better design would be to make it a regular function called `equal` so that it can take a "tolerance" parameter. It could be implemented like this:

    bool 
    equal(ITensor const& T1, ITensor const& T2, Real tol = 1E-12)
    {
    auto n1 = norm(T1);
    auto n2 = norm(T2);
    auto scal = (n1 < n2 ? n1 : n2);
    return norm(T1-T2) <= scal*tol;
    }

lfrahm:
I really like your idea of having an `equal` function with an optional tolerance parameter. It would be great if we can build it like that. I also like your very simple implementation, actually I think mine is more memory efficient, but I don't think that's going to be an issue. I kinda have overthought that. 

Are you going to implement it or do you want me to do it?

emstoudenmire:
Hi, so if you add yours and do a pull request I'd be happy to look it over. Can you just define it for ITensors and put the code in the itensor.cc file (with a declaration in itensor.h)? Then when we converge on a design I will make an IQTensor version too.

lfrahm:
I have a student testing whether my idea is actually faster. My approach works on the ITData below the tensors, I think that's overkill, so we probably go with your idea. Do you want me to put this together and prepare a pull request?

PR #117, opened 2017-03-01
also build shared library

kyungminlee:


emstoudenmire:
Looks like a really nice patch, and it is something we need. I had experimented with this before and using a dynamic lib helps significantly with the speed of compiling driver codes, as it dramatically speeds up the linking step on my machine. Let me try out your patch and then I'll pull it in.

kyungminlee:
This patch is almost the same as what is used in the hombrew formula for [itensor](https://github.com/Homebrew/homebrew-science/blob/master/itensor.rb), and I haven't had any problems so far.
The hombrew formula also creates `platform.h` so that `-DPLATFORM_xxxx` is not needed every time you compile code that uses itensor, especially if the `libitensor.dylib` is already linked to a specific LAPACK library. What do you think?

emstoudenmire:
Thanks for the patch - just pulled it. I put a flag in options.mk.sample called _ITENSOR_MAKE_DYLIB that keeps the new itensor/Makefile targets from being enabled unless users update their options.mk files. In hindsight I should have put the upper, user configurable, part of the options.mk file and the lower part which defines a bunch of flags into separate files, with the one defining flags tracked by git. That's a change I'll make in the future but it will break people's setups so I'll have to do that in a major version update.

Also I liked your platform.h idea from the homebrew formula. So I implemented that but named the file itensor/config.h. Just letting you know so you can update the formula accordingly when I release a new numbered version which I plan to do soon.

PR #119, opened 2017-05-08
Fixes Sweeps Bug and Implemented Read/Write for QCombine

lfrahm:
Hi, 
I was able to fix the problem I was describing here: http://itensor.org/support/489/segmentation-fault-when-using-combiner

I implemented the necessary read/write methods and now it works. I tried to be as consistent with your style as possible. I hope you can merge it into your repository. 

I also fixed another problem with the read/write of Sweeps. 

If you have questions or comments, just let me know. 

Best,
Lars

### DIFF
--- a/itensor/itdata/qcombiner.cc
+++ b/itensor/itdata/qcombiner.cc
@@ -17,6 +17,26 @@ namespace itensor {
 const char*
 typeNameOf(QCombiner const& d) { return "QCombiner"; }
 
+void 
+read(std::istream& s, QCombiner & dat)
+    {
+    Range range;
+    itensor::read(s,range);
+
+    using storage_type = typename QCombiner::storage_type;
+    storage_type store;
+    itensor::read(s,store);
+
+    dat = QCombiner(std::move(range),std::move(store));
+    }
+
+void
+write(std::ostream& s, QCombiner const& dat)
+    {
+    itensor::write(s,dat.range());
+    itensor::write(s,dat.store());
+    }
+
 Cplx
 doTask(GetElt<IQIndex> const& g, QCombiner const& c)
     {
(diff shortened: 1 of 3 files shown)


emstoudenmire:
Hi Lars,
Really glad you found how to fix your issue and I'd like to merge this. One question before I do though: what was the issue with the Sweeps read/write code? Was the use of auto breaking things?

emstoudenmire:
Thanks for the patch. I just merged it. Major oversight on my part that QCombiner storage didn't have read/write support! I put in a slightly newer implementation of the Sweeps read/write code

lfrahm:
Hi Miles,
Thanks for merging and updating the read/write code for Sweeps. I had the issue that after reading Sweeps from binary I got segmentation faults. Your update fixes that too. Thanks. 

By the way, big thanks for the MPI support you added recently. I think that's an awesome new feature. I extended (will extend further) your code a little bit here: https://github.com/lfrahm/MOL/blob/master/src/parallel_extended.h If you are interested I can prepare a pull request for that too. 

I have another problem with the combiner. Unfortunately I am unable to pin it down to an easy reproducible example yet. But I keep investigating it and will let you know as soon I can clearly state what the problem is. 

Thanks again, 
Lars

emstoudenmire:
Hi Lars,
Glad the MPI code is helpful. I found that it was a much needed wrapper around the MPI C interface which is itself pretty nice, but benfited from taking advantage of C++ templates and also from some logic about breaking up data into smaller chunks when sending and receiving.

I'd be glad if you prepare a pull request for those additional Environment methods. What does scatterVector do by the way?

A colleague also filed a bug report with me about IQTensor combiners. So I will look into that. Perhaps it is the same as the one you are finding? I don't know but please let me know more about the bug you are seeing.

PR #120, opened 2017-05-11
Added MPI scatterVector, sum and allSum methods.

lfrahm:
Hi Miles, 
here are the added MPI methods. I hope they are useful. 

The scatterVector method scatters a vector evenly among all processes. For example, if root has a vector with 1000 elements and there are 4 processes, scatterVector will distribute 250 elements for each process. So after calling scatterVector, all process (including root) will have a vector of 250 elements with their specific part of the original vector. 

Not sure whether you want to keep the sum(double) method.

Feel free to ask for more, or I will add them whenever I need new ones. 

I will send you an email explaining my problem with the combiner later.

Best,
Lars

### DIFF
--- a/itensor/util/parallel.h
+++ b/itensor/util/parallel.h
@@ -47,37 +47,14 @@ class Environment
     // Communication and flow control
     //
 
-    template <class T>
-    void 
-    broadcast(T& obj) const;
-
-    template <class T, class... Rest>
-    void 
-    broadcast(T & obj, Rest &... rest) const;
-
     void 
     broadcast(std::stringstream& data) const;
-
-    template <typename T>
-    void
-    scatterVector(std::vector<T> &v);
 
     void 
     barrier() const { MPI_Barrier(MPI_COMM_WORLD); }
 
     void 
     abort(int code) const { MPI_Abort(MPI_COMM_WORLD,code); }
-
-    double 
-    sum(double r) const;
-
-    template <typename T>
-    T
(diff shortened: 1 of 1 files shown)


emstoudenmire:
Hi Lars,
Thanks these would be good to have. My main comment is about whether they need to be class methods of the Environment object. Even though I made some similar methods to be class methods, for a few of them that was not necessary and I probably shouldn't have designed them that way, so I may deprecate those and replace them with free functions.

Do any of the new methods you added use the internal data members of the environment object? If not could we make them all to be free functions?

lfrahm:
Hi Miles,

Thanks for your comment. No they don't use any private data. They just use nnodes_ and rank_ and you can access these by functions too. So you are right, we could put them in free functions.

emstoudenmire:
Great, so if you don't mind changing them to free functions, possibly taking the Environment object as their first argument, I'd be happy to merge the pull request.

lfrahm:
Ok, the newly added and the broadcast functions of Environment are now free functions you can call via `broadcast(envobj, obj)`. I didn't touch the Mailbox functions, however I think `send(obj)` and `receive(obj)` can use this polish too. Let me know if you want me to add those. 

Best,
Lars

emstoudenmire:
Hi Lars,
I made a few minor changes, just putting back a couple of pre-existing class methods for backward compatibility. But I agree with preferring to use the free function version of broadcast from now on. 

Best regards,
Miles

PR #121, opened 2017-05-23
checkOrtho and exactApplyMPO

ybarlev:
Corrects unimplemented method in checkOrtho and a memory leak during instantiation in exactApplyMPO


emstoudenmire:
Thanks, these look like very important fixes!

PR #123, opened 2017-06-03
Adds thread safety for index hash generator

ybarlev:
The Generator is saved per thread and not globally, does avoiding collisions in multithread applications.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,14 +19,17 @@ configure:
 	@echo Configure: Writing current dir to this_dir.mk
 	@echo THIS_DIR=`pwd` > this_dir.mk
 	@echo "#ifndef __ITENSOR_CONFIG_H" > itensor/config.h
-	@echo "#define __ITENSOR_CONFIG_H\n" >> itensor/config.h
+	@echo "#define __ITENSOR_CONFIG_H" >> itensor/config.h
+	@echo "" >> itensor/config.h
 	@echo "#ifndef PLATFORM_$(PLATFORM)" >> itensor/config.h
 	@echo "#define PLATFORM_$(PLATFORM)" >> itensor/config.h
 	@echo "#endif" >> itensor/config.h
+	@echo "" >> itensor/config.h
 	@echo "#ifndef __ASSERT_MACROS_DEFINE_VERSIONS_WITHOUT_UNDERSCORES" >> itensor/config.h
 	@echo "#define __ASSERT_MACROS_DEFINE_VERSIONS_WITHOUT_UNDERSCORES 0" >> itensor/config.h
 	@echo "#endif" >> itensor/config.h
-	@echo "\n#endif " >> itensor/config.h
+	@echo "" >> itensor/config.h
+	@echo "#endif " >> itensor/config.h
 
 clean:
 	@echo "Removing temporary build files"
--- a/itensor/Makefile
+++ b/itensor/Makefile
@@ -75,15 +75,19 @@ libitensor-g.a: mkdebugdir $(GOBJECTS)
 	@ranlib $(ITENSOR_LIBDIR)/libitensor-g.a
 
 libitensor.$(DYLIB_EXT): $(OBJECTS)
-ifdef _ITENSOR_MAKE_DYLIB
+ifdef ITENSOR_MAKE_DYLIB
+ifeq ($(ITENSOR_MAKE_DYLIB),1)
 	@echo "Building dynamic library $(ITENSOR_LIBDIR)/libitensor.$(DYLIB_EXT)"
 	@$(CCCOM) $(DYLIB_FLAGS) -o $(ITENSOR_LIBDIR)/libitensor.$(DYLIB_EXT) $(OBJECTS) $(BLAS_LAPACK_LIBFLAGS)
 endif
+endif
 
 libitensor-g.$(DYLIB_EXT): mkdebugdir $(GOBJECTS)
-ifdef _ITENSOR_MAKE_DYLIB
+ifdef ITENSOR_MAKE_DYLIB
+ifeq ($(ITENSOR_MAKE_DYLIB),1)
 	@echo "Building dynamic library $(ITENSOR_LIBDIR)/libitensor-g.$(DYLIB_EXT)"
 	@$(CCCOM) $(DYLIB_FLAGS) -o $(ITENSOR_LIBDIR)/libitensor-g.$(DYLIB_EXT) $(GOBJECTS) $(BLAS_LAPACK_LIBFLAGS)
+endif
 endif
 
 touch_all_headers:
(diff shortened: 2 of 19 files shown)


emstoudenmire:
Hi Yevgeny,
Thanks for looking into this. It appears to be a nice change. I am reading up on these features since I wasn't familiar with them.

One question I have at the moment: I understand that the thread_local keyword makes each thread have its own separate instance of a static variable. So if the program immediately begins in a multi-threaded way I can see how each thread would be independently generating its own index IDs. But my question is what happens when the program is mostly operating in a single thread then one launches multiple threads just for a small portion of the program? I suppose then the original thread remains and the new threads get their own instances of the index generator? And they will always have distinct thread id's as guaranteed by the std::this_thread::get_id() I suppose. 

Your idea of putting in a std::hash around the thread id is also I good one.

ybarlev:
Hi Milles,
As much as I understand you are right. The parent thread will have its own
instance and the children will have their own, since they will have
different thread ids. But to be frank I am not an expert either, I was
following the solution here
<https://stackoverflow.com/questions/21237905/how-do-i-generate-thread-safe-uniform-random-numbers>,
since
in my initial attempts to parallelize a TEBD code I have run into multiple
Index collisions. It is also not clear to me, yet, whether this fix is
sufficient to make the package thread-safe.

emstoudenmire:
Ok that's helpful to know. Would you mind doing a little code experiment so we can verify the behavior? The experiment would just be to make a static thread_local int that is just set equal to the thread id of each thread and then have them all print them out, including the parent thread. Then we should see that if there is one parent thread and, say, 3 child threads the printout values 0,1,2,3 in a random order.

ybarlev:
Just made experiment using this simple code. 
`#include <iostream>
#include <omp.h>
#include <vector>
#include <thread>

int main() {

    static thread_local std::thread::id parent(std::this_thread::get_id());
    std::cout << "Parent=" << parent << std::endl;

#pragma omp parallel default(none) shared(std::cout)
    {
        static thread_local std::thread::id child(std::this_thread::get_id());
#pragma omp critical
        {
        std::cout << "Num threads=" << omp_get_num_threads() << std::endl;
        std::cout << "Child=" << child << std::endl;
        }
    }

    return 0;
}`

Bellow are the results:
`Parent=140150592331904
Num threads=3
Child=140150592331904
Num threads=3
Child=140150562330368
Num threads=3
Child=140150570723072`

Namely the ID of the parent is reused in one of the children threads. I think potentially this could lead to some problems. When the parent and child access the static variables in the same std::clock(). One care reduce the chance of collision by using a higher resolution clock...

emstoudenmire:
Thanks for doing the experiment. Actually, perhaps the experiment is a success. Instead of having the picture of a parent thread launching child threads and waiting for them to end, it seems the correct picture is rather than of the original thread creating "peer" threads and then participating as one of them. So as long as in the parallel section each thread has a distinct id, even if one of the ids is the same as in the serial section, then there should be no problem, do you agree? Let me know if you see a way this could lead to a problem but I think it is ok. (I.e. I don't think there would be a race condition that would lead to different threads both creating the same index id number.)

ybarlev:
No problem. I don't think there are race conditions, but there are
potentially index duplicates. For simplicity imagine that the generator is
created only from the hash of the id and not the clock. Then if the parent
creates an index with its generator, and in the parallel section it
"converts" itself into a child with same thread id, then it will create the
same generator as the parent, and therefore the same Index object.
Potentially this can result in contractions in unexpected places. The clock
helps mitigating this problem a bit, but due to its limited resolution
theoretically there can still be duplicates if the time which passes from
the first Index generation in the parent and the first index generation in
the child is sufficiently short. This is why I suggested to use a high
resolution clock, to reduce the probability of this event even further.

emstoudenmire:
Oh, I see. Then a key question is whether the parent thread actually does convert itself into a child thread or not. My guess is that it actually doesn't, but rather the child, or peer threads get created and the parent thread just continues along as it was before. In particular, I am thinking this means that static variables in the parent thread won't get re-constructed when entering the parallel section. I guess a simple way to check this would indeed be to run your test code but with the static variable being a call to the clock rather than the thread id. Then we could see if the static time/clock variable in the parent thread stays the same as it was in the serial section or gets re-initialized when entering the parallel section.

ybarlev:
Actually it does get reinitialized in the parallel section, see below. So
both parent and child with same thread id potentially can end up with the
same generator....

Parent=139688853303488
Parent clock=16686
Num threads=3
Child=139688853303488
Child clock=20804
Num threads=3
Child=139688831694592
Child clock=22857
Num threads=3
Child=139688823301888
Child clock=24913

emstoudenmire:
Thanks for checking. Could you please post the sample code actually? I think I may see a resolution to the issue: note that in the first sample code you posted there are two *different* static variables named parent and child. So if in your other test they are also different then it is only natural that the new child static variables will get initialized, but it's not really a second initialization of the first static variable since they are just totally distinct variables.

ybarlev:
Here is the code, and below is a more verbose output.

#include <iostream>
#include <omp.h>
#include <vector>
#include <thread>

int main() {

    static thread_local std::thread::id id(std::this_thread::get_id());
    static thread_local double clock(std::clock());
    std::cout << "Parent=" << id << std::endl;
    std::cout << "Parent clock=" << clock << std::endl;

    std::cout << "============= Parallel region ============= " <<
std::endl;
#pragma omp parallel default(none) shared(std::cout)
    {
        static thread_local std::thread::id id(std::this_thread::get_id());
        static thread_local double clock(std::clock());
#pragma omp critical
        {
        std::cout << "Num threads=" << omp_get_num_threads() << std::endl;
        std::cout << "Child=" << id << std::endl;
        std::cout << "Child clock=" << clock << std::endl;
        }
    }

    std::cout << "============= End parallel region ============= " <<
std::endl;

    std::cout << "Parent=" << id << std::endl;
    std::cout << "Parent clock=" << clock << std::endl;

    return 0;
}

Output
=====

Parent=140566938068992
Parent clock=40675
============= Parallel region =============
Num threads=3
Child=140566938068992
Child clock=42605
Num threads=3
Child=140566908045056
Child clock=42606
Num threads=3
Child=140566916437760
Child clock=42608
============= End parallel region =============
Parent=140566938068992
Parent clock=40675

martenrichter:
Hi,
is there any news about solving the multithreading problem?
I will probably soon run into the same problem.
I think having a separate random generator for every thread is probably a good idea.
Regarding the seed, I do not know if taking just the time and id is really working well.
I had in the past very good results using the /dev/random and /dev/urandom devices for getting the initial seed, which is purely random, but also very slow.
What do you think about the index object having a pool of random generators, 
where it is made sure, that every generator is only accessed by one thread at a time.

Marten

ybarlev:
Hi Marten,
Using the hashed id as the seed is working for me. I didn't notice any
collisions...
Best,
Y.

emstoudenmire:
Thanks again for this bug fix, and for your patience as I took forever to include it. I just included it in commit a5305e49101d0

PR #124, opened 2017-06-07
Read/Write for Args and new MPI method

lfrahm:
Hi Miles, 
here comes the implementation and test of the new read/write method for Args. I think I followed your very helpful instructions quite closely. Thanks! 

One thing that surprised me is that the read/write methods for InfArray are in `itensor/util/readwrite.h` and not in `#include "itensor/util/infarray.h`, therefore I had to include the readwrite.h in the args.cc. But that is up to you.

I also added an MPI gatherVector method that collects the vector elements of all procceses to the root node. I hope you like it.

Let me know if you have questions or comments!
Best,
Lars

### DIFF
--- a/itensor/util/args.cc
+++ b/itensor/util/args.cc
@@ -7,6 +7,7 @@
 #include <iostream>
 #include "itensor/util/args.h"
 #include "itensor/util/error.h"
+#include "itensor/util/readwrite.h"
 
 namespace itensor {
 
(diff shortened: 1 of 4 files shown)


emstoudenmire:
Great contributions, thanks. The reason that the InfArray read write is in readwrite.h is that I first created a read and write for std::vector there, then put the InfArray one in the same place. But I agree probably it would make more sense in uti/linfarray.h

PR #128, opened 2017-08-01
enclose error funcs in itensor namespace

mileschen360:
to avoid name conflicts to other libraries such as boost

### DIFF
--- a/itensor/util/error.h
+++ b/itensor/util/error.h
@@ -9,6 +9,7 @@
 #include <stdexcept>
 #include <iostream>
 
+namespace itensor{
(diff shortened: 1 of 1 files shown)


emstoudenmire:
Yes, this is definitely needed. Thanks for catching this mistake! (These error functions date back to a much older version of the library before we enclosed things in a namespace.)

emstoudenmire:
I added your name to the list of contributors here: http://itensor.org/about.html
Let me know if I got your information correct, and let me know if you want me to put your institution or link to something other than your Github page.

PR #132, opened 2017-08-27
Fixed Makefile's to handle paths properly.

kyungminlee:
The directories for the target library files (`libitensor.a`, etc) need to be specified.


emstoudenmire:
Good catch. Thanks!

PR #134, opened 2017-08-28
Moved local static variables in global.h to global.cc

kyungminlee:


emstoudenmire:
Thanks for a good pull request. Sorry for the slow action on it!

PR #140, opened 2017-12-14
Add order() command to force index order of an ITensor

mtfishman:


emstoudenmire:
Thanks for the new feature Matt. Looks good. I did some testing and everything works. Please do add the new .set (with no IndexVals) since that was one of the main features I was using with the older "ordered" function.

PR #147, opened 2018-01-17
order: "..." syntax

### LINKED ISSUE #144: order: "..." syntax
Add a "..." syntax to the order method that is a place holder for indices that the user doesn't care about the order of.

mtfishman:
Adds "..." syntax for order so that not all indices need to be specified. Right now, only supports "..." as the first or last input (closes #144).

### DIFF
--- a/itensor/itensor_interface.h
+++ b/itensor/itensor_interface.h
@@ -106,16 +106,13 @@ class ITensorT
     set(IV const& iv1, VArgs&&... ivs)
         -> stdx::if_compiles_return<void,decltype(iv1.index),decltype(iv1.val)>;
 
-    template<typename... VArgs>
+    template<typename Int, typename... VArgs>
     auto
-    set(int iv1, VArgs&&... ivs)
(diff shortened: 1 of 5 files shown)


emstoudenmire:
Hi Matt, great feature add. I merged in the code, but made some small changes:

- I changed the second argument of detail::getInds to just be std::string& since that's the only way it gets called. I figured this could prevent any bugs.

- I renamed getInds to getDotInds just to clarify that's what it's used for

- Note that when using formatting strings, %d formats ints and %s formats strings (you used a %d to format a string).

- I changed some for(auto I : inds) loops to for(auto& I : inds). Even though Index and IQIndex objects are designed to be cheap to copy, I don't think it would be any slower to loop over them by reference, and it could be faster (say if IQIndices become more expensive to copy later if their design is changed later on). But for simple "pod" (plain old data) types like ints it's still a good idea just to copy and for(auto i : range(N)) would still be appropriate for example.

PR #151, opened 2018-02-28
Extend .set to work with long int

mtfishman:
I noticed with the following type of code:
```
Index i("i",2), j("j",2);
auto A = randomTensor(i,j);
for(auto s : range1(i.m())) A.set(s,s,sqrt(A.real(s,s)));
```
the new .set, .cplx, etc. aren't dispatching correctly, since the new versions were limited to just accepting `int`'s while `s` above becomes type `long int`, so it tries to use the versions that are meant for `IndexVal`'s and throws an error. I extended .cplx and .set to accept anything that is convertible to an `int`, let me know if this seems like the way to go.

### DIFF
--- a/itensor/itensor_interface.h
+++ b/itensor/itensor_interface.h
@@ -103,14 +103,14 @@ class ITensorT
     template <typename IV, typename... IVs>
     auto
     cplx(IV const& iv1, IVs&&... ivs) const
-        -> stdx::enable_if_t<std::is_same<IV,IndexVal>::value
-                             || std::is_same<IV,IQIndexVal>::value,
+        -> stdx::enable_if_t<std::is_same<IV,IndexVal>::value || std::is_same<IV,IQIndexVal>::value,
                              Cplx>;
 
     template <typename Int, typename... Ints>
     auto
     cplx(Int iv1, Ints&&... ivs) const
-        -> stdx::enable_if_t<std::is_convertible<Int,int>::value,Cplx>;
+        -> stdx::enable_if_t<std::is_same<Int,int>::value || std::is_same<Int,long int>::value,
+                             Cplx>;
 
     Cplx
     cplx() const;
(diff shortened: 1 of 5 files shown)


emstoudenmire:
Great - I had noticed this bug too and was planning to fix but you beat me to it. 

Before I pull the changes, can you please also include a unit test that reproduces the bug (to guarantee it can't happen again after further changes to the library)?

mtfishman:
Added unit tests. In the first commit I was using:
```
stdx::enable_if_t<std::is_convertible<Int,int>::value,void>
```
to dispatch on the `int`'s version of `.set` and `.cplx`, but I noticed that meant that inputs like:
```
Index i("i",2), j("j",2);
auto A = randomTensor(i,j);
A.set(2.3,1,2);
```
wouldn't throw an error. I changed it to:
```
stdx::enable_if_t<std::is_same<Int,int>::value || std::is_same<Int,long int>::value,void>
```
which seems to have fixed that. Are there other kinds of `int`-like inputs we should account for?

itensor/itensor_interface.h
```diff
@@ -117,12 +121,14 @@ class ITensorT
     template<typename IV, typename... VArgs>
     auto
     set(IV const& iv1, VArgs&&... ivs)
-        -> stdx::if_compiles_return<void,decltype(iv1.index),decltype(iv1.val)>;
```
mtfishman: Is this ok to remove? I'm not sure what it is doing.

emstoudenmire: Hi Matt,
Let's leave that unless it's causing an issue. Rule #1 of editing a library is that if you don't know what a line of code is doing, it's probably some bug fix for a rare weird case that happened once long ago. (Applies also to code you yourself wrote then forgot why.)

What that line looks to be doing is making sure that "IV" is either the IndexVal or IQIndexVal type, since those types have a ".index" and ".val" field.

emstoudenmire:
Good catch about not using is_convertible. That actually made me a bit nervous and I should have mentioned it, because a lot of things may be convertible to int. 

So instead of making separate trait tests just for int and long, how about the "is_integral" trait. http://en.cppreference.com/w/cpp/types/is_integral

I think that's the right way to go, but would need to test to make sure.

itensor/itensor_interface.h
mtfishman: Definitely a good rule. I changed it back, I thought I had to change it to get it to dispatch properly on the different versions of .set but it is working properly with it there.

mtfishman:
I changed over to using is_integral. That is exactly what I was looking for, thanks! I think .set, .cplx, etc have all the behavior we want now.

emstoudenmire:
Great. Ok merged!

emstoudenmire:
Actually there was an issue getting the latest changes to compile on my machine. The "order" method definitions in itensor_interface.h weren't matching the ones in intensor_interface.ih. So I just pushed a patch.

PR #154, opened 2018-03-14
Minor fixes

kyungminlee:
- `indexset.ih`
Return an empty local `static const IndexT` instance instead of creating one every time. This fixes warnings since `findtype` and `finddir` both return lvalue references.
- `qn.h`
Defined unary minus for `const QNVal`

### DIFF
--- a/itensor/indexset.ih
+++ b/itensor/indexset.ih
@@ -505,7 +505,7 @@ dir(const IndexSetT<IndexT>& is, const IndexT& I)
 
 
 template <class IndexT>
-const IndexT&
+IndexT
 finddir(const IndexSetT<IndexT>& iset, Arrow dir)
     {
     for(const auto& J : iset)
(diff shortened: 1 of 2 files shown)


emstoudenmire:
Hi Kyungmin,
Thanks for the patches. The issue with findtype and finddir should definitely be fixed. However, I think the better solution is to change the return type to just be IndexT instead of const IndexT&. The reason is that it was only a "micro optimization" in the first place, and I want to be careful about where and when static variables are used (they can inhibit thread safety, etc.).

emstoudenmire:
I've been pretty behind on pull requests by the way. I'll make sure to take a look at your other ones. Thanks for those -

kyungminlee:
@emstoudenmire I agree. Especially the unary minus on QNVal& is a bad idea IMO.

emstoudenmire:
Yes, I agree about the unary minus. Not sure why I defined it that way. We should just replace the QNVal& method by the new one you added. Can you please update the patch to make both changes (return by value findtype and finddir and return by value unary minus)? Thanks

kyungminlee:
Done

Issue #155, opened 2018-05-11
Build fails because of undefined shared_ptr


kyungminlee:
The build on Linux with g++ fails at `itdata/dense.cc` which includes `iqindex.h` because of the undefined `shared_ptr`.

https://gist.githubusercontent.com/kyungminlee/398508f141f5c68e8a474c4b28a1ef36/raw/a558fa5c1d186bae526c799b81914e47d9d482f5/build-itensor.log

PR #156, opened 2018-05-11
Added `#include <memory>` to global.h

kyungminlee:
Fixes issue https://github.com/ITensor/ITensor/issues/155


emstoudenmire:
Thanks - I actually had done this too but for some reason it didn't end up in the latest version of the code!

PR #160, opened 2018-07-08
Fix catching polymorphic `TooSmallForReal` by value

mwallerb:
`itensor::TooSmallForReal` is derived from `std::runtime_error`.  This is a
polymorphic type, so cathing it by value is dangerous.  gcc 8 warns about
this:

    [...]/itensor/itensor/itensor_interface.ih:220:5: warning:
    catching polymorphic type ‘class itensor::TooSmallForReal’ by value
         catch(TooSmallForReal)
         ^~~~~

This fixes this bug.


emstoudenmire:
Great! Thanks for catching this & for the contribution.

PR #161, opened 2018-08-20
AutoMPO for iDMRG

akeselman:
Adjusting AutoMPO implementation to work for infinite systems.

When specifying the Hamiltonian, terms that connect a site in one unit cell to a site in the next unit cell should be specified with site indices (i, N+j), where i and j are the site indices within a unit cell and N is the number of sites in a unit cell.

Note that AutoMPO does not perform any compression in this case - when constructing the finalMPO, the SVD is skipped and the matrix elements of the MPO are taken directly from tempMPO.


emstoudenmire:
Thanks a lot Anna! We (me and Matt Fishman) will be looking into this and working on incorporating it into ITensor -

krvajal:
Any chance of this been merged?

PR #164, opened 2018-10-09
Add GMRES with restarts

mtfishman:


emstoudenmire:
Looks great at a first pass. See above though: it looks like it's not compiling on one of the Jenkins test systems. I think it has to do with line 457 of eigensolver.h

mtfishman:
Just a silly mistake, I didn't realize std:conj() always outputs complex, even when you input a real. Strange that it worked fine with the older compiler.

emstoudenmire:
Well written code & with a unit test. Great! Nice to have this in the library.

PR #166, opened 2018-10-12
Add applyMPO method which allows for a starting guess wavefunction

mtfishman:
1. I added an extra applyMPO method where a user can input a starting guess for the output wavefunction (as we discussed, this only works for "Fit" and throws an error for "DensityMatrix". Also, the guess wavefunction is not overwritten)
2. The "DensityMatrix" method did not have an option to normalize the output wavefunction, so I added that for consistency with the "Fit" method
3. I changed the default of applyMPO to not normalize the output wavefunction

### DIFF
--- a/itensor/mps/mpo.h
+++ b/itensor/mps/mpo.h
@@ -288,6 +288,13 @@ applyMPO(MPOt<Tensor> const& K,
          MPSt<Tensor> const& x,
          Args const& args = Args::global());
 
+template<class Tensor>
+MPSt<Tensor>
+applyMPO(MPOt<Tensor> const& K,
+         MPSt<Tensor> const& x,
+         MPSt<Tensor> const& x0,
+         Args const& args = Args::global());
+
 //
 // Applies an MPO to an MPS using the zip-up method described
 // more fully in Stoudenmire and White, New. J. Phys. 12, 055026 (2010).
--- a/itensor/mps/mpoalgs.cc
+++ b/itensor/mps/mpoalgs.cc
@@ -97,11 +97,17 @@ applyMPO(MPOt<Tensor> const& K,
     {
     auto method = args.getString("Method","DensityMatrix");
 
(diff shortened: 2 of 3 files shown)


emstoudenmire:
Looks good. I guess I had already put in checkMPOProd? Or did you add that in a previous commit?

emstoudenmire:
I like the approach of making applyMPO have consistent normalization behavior but leaving fitApply as-is for now. But we might want to make a breaking change to fitApply in the future to make it not normalize by default (and maybe even remove that as an option altogether).

mtfishman:
> Looks good. I guess I had already put in checkMPOProd? Or did you add that in a previous commit?

If you are asking if checkMPOProd was already in ITensor, it was already there (I didn't put it in) but I only realized it was there after the first pull request was merged so I decided to use it here.

> I like the approach of making applyMPO have consistent normalization behavior but leaving fitApply as-is for now. But we might want to make a breaking change to fitApply in the future to make it not normalize by default (and maybe even remove that as an option altogether).

I figure this is a good intermediate solution. It seems like we should just be moving toward removing fitApplyMPO, exactApplyMPO, etc. anyway. My next step is to figure out the right way to deprecate those old functions. I am leaning towards adding warnings to them that point people to applyMPO. If people are getting a ton of warnings, it may inspire them to switch it in their code, but it is a bit nicer than just immediately throwing errors.

emstoudenmire:
One thing I’d suggest is just “documentation deprecating” them for now (hiding them from the documentation, which I think maybe we’ve already done?). Then also let’s move them to the end of the mpo.h header file and place comments around them saying please use applyMPO instead. Then yes we can also add a warning. I was over-thinking things when I mentioned the thread safety of the warning count the other day; worst case I think if multithreaded it would probably just print out more than one warning, no big deal. So I’d suggest putting a static integer to count the warning number and just print the warning once.

PR #167, opened 2018-10-15
Implementation of LU and Cholesky solvers for linear systems

jurajHasik:
An initial implementation for LU and Cholesky solvers for linear systems. The linsystem function 
takes Tensor A, interprets it as a matrix of coefficients and solves for the right hand side passed as tensor B.
A choice for the solver, method="LU" or "Cholesky", is passed in through Args system with default being LU.

### DIFF
--- a/itensor/Makefile
+++ b/itensor/Makefile
@@ -10,7 +10,6 @@ include ../options.mk
 SOURCES+= util/args.cc     
 SOURCES+= util/input.cc
 SOURCES+= util/cputime.cc
-
 SOURCES+= tensor/lapack_wrap.cc 
 SOURCES+= tensor/vec.cc 
 SOURCES+= tensor/mat.cc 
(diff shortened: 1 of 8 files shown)


mtfishman:
Hi, thanks for the pull request. You may have noticed that I recently added an implementation of gmres, which is an iterative solver for linear systems. In general in tensor network calculations, I think it is better to use something like gmres/bicgstab, since you can take advantage of the sparse nature of the tensor network (i.e. for solving A*x = b for x, one does not need to form the matrix A, but instead just needs the result of acting A on a vector, which in tensor network calculations is usually much more efficient since the linear map A is usually made up of the contraction of multiple tensors). 

Having an alternative based on an LU/Cholesky decomposition may be useful as well. Did you have a particular use case where something like gmres() would not be sufficient?

Also, do you have an implementation for the case with IQTensors?

PR #169, opened 2018-10-19
inline fix for GMRES

mtfishman:


emstoudenmire:
Hi Matt, see above - looks like one of the Jenkins builds is failing still?

mtfishman:
The failure looks unrelated to this commit, it is actually the same SVD test that has failed randomly before. I think we should just relax the bound on that test (I added a new commit in the gmresfix branch but for some reason it is not showing up in this pull request). 

Also, maybe before merging we should have @kshyatt test this on her code (I did a quick test on code she sent me, and it seemed to fix the error, but just do double check).

emstoudenmire:
Yes, I think it's reasonable to relax the thresholds on those numerical tests. These hardly ever fail on my machine(s) but for some reason they are failing a lot on Jenkins. (Now another one - not SVD - is failing.)

PR #171, opened 2018-11-01
Add more constructors for diagTensor() and delta()

mtfishman:
(To create consistency with ITensor constructors.)

### DIFF
--- a/itensor/iqtensor.h
+++ b/itensor/iqtensor.h
@@ -87,10 +87,23 @@ combiner(IQIndex const& i1,
 
 //Construct diagonal IQTensor with diagonal 
 //elements set to 1.0
+IQTensor
+delta(IQIndexSet const& is);
+
 template<typename... Inds>
 IQTensor
 delta(IQIndex const& i1,
       Inds const&... inds);
+
+IQTensor
+delta(std::vector<IQIndex> const& is);
+
+template<size_t N>
+IQTensor
+delta(std::array<IQIndex,N> const& is);
+
+IQTensor
+delta(std::initializer_list<IQIndex> is);
 
 IQIndex
 findIQInd(IQTensor const& T, Index const& i);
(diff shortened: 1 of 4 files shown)


emstoudenmire:
Hi Matt,
One idea to improve this could be to just make the delta(IQIndexSet const& is); function signature be delta(IQIndexSet is); instead. Then, because IQIndexSet can be constructed from array, initializer_list, etc I think those cases may then automatically work without requiring extra code. The slight downside is that it will cause a copy for the case that the argument really is already just an IQIndexSet. But I think this will be a very slight extra cost, if any. What do you think?

mtfishman:
I don't have a strong opinion either way (and it is definitely a bit annoying defining many functions by hand so a more general approach would be nice). 

However, I just tried a test where I changed the signature to delta(IndexSet is), removed the version explicitly taking a vector, and ran the following:

```
auto i = Index("i",2);
auto j = Index("j",2);
std::vector<Index> inds = {i,j};
auto I = delta(inds);
```

and I get an error "no matching function for call to 'delta(std::vector<itensor::Index, std::allocator<itensor::Index> >&)'". This is followed by: "note:   no known conversion for argument 1 from 'std::vector<itensor::Index, std::allocator<itensor::Index> >' to 'itensor::IndexSet {aka itensor::IndexSetT<itensor::Index>}'". It is weird, because you can definitely call IndexSet(inds) to convert to an IndexSet (i.e. delta(IndexSet(inds)) works as expected). Maybe the conversion is being done in a way that C++ doesn't recognize it?

Anyway, I am inclined to just keep it as is, and if we find there are too many functions later on we can investigate a more generic solution based on conversions.

emstoudenmire:
I see - yes this is because that IndexSet constructor is marked explicit, so requires one to explicitly write IndexSet(v) where v is a std::vector. I could remove the explicit keyword, but maybe that will lead to too much "magic" i.e. unexpected conversions in other places. So I agree, let's go with the pull request as it is; it's already very reasonable, I was just trying to improve.

PR #178, opened 2018-11-07
Have final svdBond respect args in nmultMPO

kshyatt:
If you don't do this, the "last" bond in the resulting MPO can have way (WAY) higher bond dimension than all others. Now, the svdBond respects the "global" Maxm.


emstoudenmire:
Good catch - thanks!

Issue #180, opened 2018-11-09
Comments on commit 7dca9a7


mingruyang:
This changes does not work for small blocks. 1E-3 seems to be too large, which makes docut larger than sqr(d(n)) in function svdImpl in svd.cc and results this_m=0. This makes initialization of the QN of IQIndex L and R fail. I changed it back to 1E-5 and it works for me.

PR #181, opened 2018-11-09
Make it able to catch the possible runtime error from empty IQIndex a…

mingruyang:
…fter SVD in debug mode. Refer to issue #180.


emstoudenmire:
Looks good. Thanks Mingru!

Issue #182, opened 2018-11-12
Integer overflow


mingruyang:
When the bond dimension of MPS is very large, e.g., 6000, the value of the size of LocalOp will overflow the maximum of the long integer type and thus give a undefined behavior (for my case,it gives negative value.) This will make line 107 area(phi.front().inds()) != size_t(maxsize) in iterativesolvers.h becomes true and throws an error.

mingruyang:
size_t() won't help it.

mtfishman:
Hi Mingru,

Thanks for pointing this out. What is the local site dimension of your MPO? 

Also, do both sides of the check on line 107 overflow (i.e. does either size_t(maxsize) or area(phi.front().inds()) give you the correct result for the size M^2*d^2 of your effective Hamiltonian, where M=6000 is the MPS bond dimension and d is the local site dimension of your MPO)? 

You may need to send me the code you are running to help track down where the overflow is happening.

Right now, the size_ field of LocalOp is type long, we may need to change that if the size you are seeing is larger than that. However, we may be able to avoid calculating that size in davidson anyway. 

Cheers,
Matt

emstoudenmire:
Also, just wanted to mention that as a temporary workaround you should be able to comment out this line of iterativesolvers.h and then everything else should work just fine. It's not an essential line of code, but rather just a check that the input is valid (which it will be inside of our DMRG code). 

But thanks a lot for pointing this out and we'll fix it!

mtfishman:
Yes, I meant to add that as well. Although depending on what is overflowing, maxsize is used elsewhere in davidson so if that is overflowing it could possible cause other problems (though I think the other usage can just be replaced by maxiter_/actual_maxiter).

mingruyang:
Thanks Miles and Matt!

The local site dimension d=4 (just like the Hubbard model). I use a three-site DMRG algorithm but a two-site DMRG with noise can also give such errors in my case, given a larger bond dimension. 

I tried for three times. area(phi.front().inds()) gives me the correct result while size_t(maxsize) is negative, and their absolute values always add up to 2^32= 4,294,967,296. The error messages of the three tries are as follows:
1. bond dimension of MPS = 5984
area(phi.front().inds()) = 2174486784
A.size() = -2120480512
From line 99, file /home/mingruy/itensor/itensor/iterativesolvers.h
davidson: size of initial vector should match linear matrix size
davidson: size of initial vector should match linear matrix size
Aborted (core dumped)
[...]

PR #184, opened 2018-11-13
Fix integer overflow of line 107 of iterativesolver.h. Refer to issue…

mingruyang:
… #182 and pull request #183.


emstoudenmire:
Fixed by #183 (if not please reopen).
Thanks for pointing out this issue and suggesting a fix, Mingru!

PR #185, opened 2018-11-14
Add optional rescale to checkMPOProd

ryanlevy:
When the new resulting MPS has a large norm, `checkMPOProd` slightly lies to you. While one can manually divide by the norm, it's already being calculated inside the function so you might as well save it and apply it then.

### DIFF
--- a/itensor/mps/mpo.cc
+++ b/itensor/mps/mpo.cc
@@ -557,10 +557,12 @@ template<class Tensor>
 Real
 checkMPOProd(MPSt<Tensor> const& psi2,
              MPOt<Tensor> const& K, 
-             MPSt<Tensor> const& psi1)
+             MPSt<Tensor> const& psi1,
+             bool rescale)
     {
     //||p2> - K|p1>|^2 = (<p2|-<p1|Kd)(|p2>-K|p1>) = <p2|p2>+<p1|Kd*K|p1>-2*Re[<p2|K|p1>]
     Real res = overlap(psi2,psi2);
+    Real scale = res;
     res += -2.*overlapC(psi2,K,psi1).real();
     //Compute Kd, Hermitian conjugate of K
     auto Kd = K;
@@ -569,11 +571,12 @@ checkMPOProd(MPSt<Tensor> const& psi2,
         Kd.Aref(j) = dag(swapPrime(K.A(j),0,1,Site));
         }
     res += overlap(psi1,Kd,K,psi1);
+    if (rescale) res/=scale;
     return res;
     }
 template
-Real checkMPOProd(MPSt<ITensor> const& psi2, MPOt<ITensor> const& K, MPSt<ITensor> const& psi1);
+Real checkMPOProd(MPSt<ITensor> const& psi2, MPOt<ITensor> const& K, MPSt<ITensor> const& psi1, bool rescale);
 template
-Real checkMPOProd(MPSt<IQTensor> const& psi2, MPOt<IQTensor> const& K, MPSt<IQTensor> const& psi1);
+Real checkMPOProd(MPSt<IQTensor> const& psi2, MPOt<IQTensor> const& K, MPSt<IQTensor> const& psi1, bool rescale);
 
 } //namespace itensor
--- a/itensor/mps/mpo.h
+++ b/itensor/mps/mpo.h
@@ -466,7 +466,8 @@ template<class Tensor>
 Real
 checkMPOProd(MPSt<Tensor> const& psi2,
              MPOt<Tensor> const& K, 
-             MPSt<Tensor> const& psi1);
+             MPSt<Tensor> const& psi1,
+             bool rescale=false);
 
 //
 // Deprecated interfaces - kept for backwards compatibility
--- a/unittest/mpo_test.cc
+++ b/unittest/mpo_test.cc
@@ -213,6 +213,49 @@ SECTION("applyMPO (Fit)")
 
     }
 
+SECTION("checkMPOProd Scaling")
+    {
+
+    auto method = "DensityMatrix";
+
+    auto N = 10;
+    auto sites = SpinHalf(N);
+
+    auto psi = MPS(sites);
+
+    //Use AutoMPO as a trick to get
+    //an MPO with bond dimension > 1
+    auto ampo = AutoMPO(sites);
+    for(auto j : range1(N-1))
+        {
+        ampo += "Sz",j,"Sz",j+1;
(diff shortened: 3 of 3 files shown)


mtfishman:
Thanks for the pull request! You make a good point that `checkMPOProd()` should generally be doing some rescaling according to the norm.

I think maybe we should actually go further with this and make a new function `errMPOProd()` that outputs a proper error, and leave `checkMPOProd()` as-is to possibly be deprecated or changed in the future (to me, the name `checkMPOProd()` implies that the function is doing a check, so it could possibly be something like `checkMPOProd(psi2,K,psi1,err_threshold) -> bool` that uses the error from `errMPOProd()` and a threshold to say if `|psi2>` is close enough to `K|psi1>`).

It seems like the error defined by `errMPOProd(psi2,K,psi1)` should be:
```
err = || |psi2> - K|psi1> || / || K|psi1> || = 
\sqrt{1 + (<psi2|psi2> - 2*Re[<psi2|K|psi1>]) / <psi1|K^{\dagger}K|psi1>}
```
(note that I think it makes more sense to normalize by `|| K|psi1> ||` since that is the "exact" vector we are comparing against).

@emstoudenmire, what do you think about that?

Cheers,
Matt

ryanlevy:
I went ahead and implemented the suggestion in [a new branch](https://github.com/ryanlevy/ITensor/tree/errorProd) and can submit a new PR once Miles chimes in. 

I will say that the error formula you suggested Matt does get rid of a sort of "global phase" information that the original one had (as currently `checkMPOProd` can be negative)

mtfishman:
Thanks for your work on this!

I am a bit confused, I don't see why the original `checkMPOProd()` could be negative/have a phase, since it is calculating: `|| |d> ||^2 = <d|d>` where `|d> = |psi2> - K|psi1>`, and I don't see how the length/squared length of a vector could be negative. Did you have an example of that happening in practice? Note that even though the vectors `|psi2>` and `K|psi1>` are represented as MPS's or products of MPO's and MPS's, the overlaps are still calculated exactly.

Along the same lines I am not sure if taking the absolute value in the `errMPOProd()` you implemented is strictly necessary, but maybe it is a good idea in case the difference is extremely small (i.e. near machine precision) and numerically it comes out to a very small negative number, which would be a problem for taking the square root.

ryanlevy:
I do see this in practice, but it looks like after switching the normalization to the not-approximate `K|p1>` things are sensible again (sorry about that).

I have an MPO*MPS with the following values:
```
<p2|p2>        = 2.69827e+12
-2Re[<p2|K|p1> = -5.39654e+12
<p1|KdK|p1>    = 2.69827e+12
err ~ -1.11022e-15
```
My guess is its machine precision error with small values. If I severely reduce `Maxm` the error ||...||^2 is always positive.

mtfishman:
No worries, thanks for clarifying. I just wanted to make sure that there wasn't something wrong with the implementation or something I was not understanding. In that case it seems like a good idea to take the absolute value before taking the square root (as you implemented already), in case these machine precision generated negative numbers pop up.

Out of curiosity, what is your use case where these large norms are showing up?

ryanlevy:
I have random MPOs that aren't normalized in any particular way (basically I call `generate` on each ITensor producing random numbers). I would just divide out by the norm and avoid this but sometimes they underflow to 0 and so ITensor causes a crash

ryanlevy:
To construct the MPO I did something like the following:
```c++
MPO K;
  for(int i=1;i<=N;i++){
    auto &t=K.Aref(i);
    t.generate(returnRandom); //lambda function
    //t/=norm(t); //crashes when norm(t)==0
}
```
Thanks for the suggestions! Norm for some reason isn't defined on `MPO`s  but I now realize that `overlap(K,K)` works just fine (and for small systems (10 sites) is `1E13` already!). Looks like `norm(K)^{1/N}~4` which means that it reduces `overlap(K,K)~1E11` for some concrete numbers. When I crank up system size I'm still worried about overflow though. 

I haven't noticed any problems doing `K|psi>` after normalizing, but until I changed this function it's been a struggle determining if the MPO application was working or not.

mtfishman:
I think normalizing the individual tensors like in your sample code should generally be good enough, unless you really need the MPOs properly normalized to 1 for some reason. If so, it is probably best to first normalize the MPO tensors (as you show above) before normalizing with `norm(K) = sqrt(overlap(K,K))`. Here is an example of what I mean:

```c++
auto N = 100;

//Use AutoMPO as a trick to create a random
//MPO with a finite bond dimension
//We should probably create a constructor that
//can do this directly
auto sites = SpinHalf(N);
auto ampo = AutoMPO(sites);
for(auto j : range1(N-1))
    {
    ampo += "Sz",j,"Sz",j+1;
    ampo += 0.5,"S+",j,"S-",j+1;
    ampo += 0.5,"S-",j,"S+",j+1;
    }
auto K = MPO(ampo);
for(auto j : range1(N))
    randomize(K.Aref(j));

//Initially, the norm is very large 
auto normK = std::sqrt(overlap(K,K));
PrintData(norm(K.A(2)));
PrintData(normK);

//First normalize the tensors individually
for(auto j : range1(N))
    K.Aref(j)/=norm(K.A(j));

//Now the norms are better controlled
normK = std::sqrt(overlap(K,K));
PrintData(norm(K.A(2)));
PrintData(normK);

//Now we can normalize the MPO properly
for(auto j : range1(N))
    K.Aref(j)/=std::pow(normK,1.0/N);
normK = std::sqrt(overlap(K,K));
PrintData(norm(K.A(2)));
PrintData(normK);
```

I think why we don't have a `norm(MPO)` function right now is because the norms of MPOs are not always well controlled, unlike the norms of (gauged) MPSs. Even now in ITensor, the `norm(MPS)`  function only works if the MPS is gauged properly, in which case the norm of the MPS is just the norm of the center site tensor and the norm is easy to compute (the cost is constant in the system size, once the MPS is gauged). In general, we don't work with MPOs gauged in any particular way, so a similar definition of the norm would not be very useful (and as we see, the total norm for a reasonably sized system would likely just tend quickly to infinity or zero). 

But a more general `normPerSite(MPS/MPO)` function that gets the norm of an MPS/MPO that is not in a particular gauge may be useful (a naive implementation basically being the code above, where the individual norms would need to be kept track of and it possibly outputs a vector of norms, one per site of the MPO or MPS).

ryanlevy:
@mtfishman thanks for your help, I'll look into playing with this `normPerSite`. In the meantime I'll close this since Miles approved the other PR!

PR #186, opened 2018-11-15
WIP: replace IndexType and name with Tags

mtfishman:
Here is a working version of the new `Tag` system. An `Index` now contains a `TagSet` which stores a small group of `Tag`s, which replaces both the `IndexType` and name of an `Index`. Index equality is now defined such that the Index id, prime level, and tags must be equal. Here are some examples of the functionality:

```c++
//Make some Site and Link indices
auto l = Index(3,"Link");
auto s = Index(2,"Site");

//Make some copies of the Site index with other Tags
auto sa = addTags(s,"a");
auto sb = addTags(s,"b");
auto sc = addTags(s,"c");

//Check contraction
//Indices sa and s will not contract with each other because they have different tags 
auto A = randomTensor(sa,sb,sc,l);
auto B = randomTensor(s,sb,sc,l);
auto C = A*B;

Print(A);
Print(B);
PrintData(C);
PrintData(rank(C)==2);

auto I = ITensor(s,prime(sa),sb,sc,prime(l));

Print(I);

//Find a specific Index
PrintData(findIndex(I,"Site,a",1)==prime(s("Site,a")));

//Priming methods for ITensor (these versions make a copy of the ITensor)
PrintData(prime(I));    // increase prime of all indices by 1
PrintData(prime(I,2));  // increase prime of all indices by 2
PrintData(prime(I,"Site"));  // increase prime of all indices with tag "Site" by 1
PrintData(prime(I,2,"Site,b"));  // increase prime of all indices with tags "Site,b" by 2
PrintData(prime(I,-1,"Link"));   // decrease prime of all indices with tag "Link" by 1
PrintData(prime(I,4,findIndex(I,"Site,b")));  // increase prime level of specific index by 4
PrintData(noPrime(I));  // set all prime levels to 0
PrintData(noPrime(I,"Site"));  // Set all prime levels of indices with "Site" tag to 0
PrintData(setPrime(I,1,"Link"));  // Set all prime levels of indices with "Link" tag to 0
PrintData(setPrime(I,2,"Link"));  // Set all prime levels of indices with "Link" tag and prime level 1 to 2
PrintData(mapPrime(I,1,2,"Site"));  // Set all prime levels of indices with "Link" tag and prime level 1 to 2
PrintData(swapPrime(I,0,1));       // All indices with prime level 0 become 1 and vise versa
PrintData(swapPrime(I,0,1,"Site"));   // All "Site" indices with prime level 0 become 1 and vice versa

//Tagging methods for ITensor (these versions make a copy of the ITensor)
PrintData(addTags(I,"i,j"));  // Add tags "i,j" to all indices
PrintData(addTags(I,"bra",1));  // Add tag "bra" to all indices with prime level one
PrintData(addTags(I,"i",findIndex(I,"Site,a",1)));  // Add a tag to a specified index
PrintData(addTags(I,"j,2","Site"));  // Add tags "j,2" to all indices with tag "Site"
PrintData(addTags(I,"k","Site",1));  // Add tag "k" to all indices with tag "Site" and prime level one

PrintData(removeTags(I,"Site"));  // Remove tag "Site" from all indices
PrintData(removeTags(I,"a",1));  // Remove tag "a" from all indices with prime level 1
PrintData(removeTags(I,"Site",findIndex(I,"Site,a",1)));  // Add a tag to a specified index
PrintData(removeTags(I,"Site","c"));  // Remove tag "Site" from all indices with tag "c"
PrintData(removeTags(I,"a","Site",1));  // Remove tag "a" from all indices with tag "Site" and prime level one
```

Breaking changes are the following:

- Changed the Index constructor so that the dimension must come first, i.e. `Index(3,"Link,4",2)` creates an Index of size 3, with a `TagSet` made of `Tag("Link")` and `Tag("4")` and prime level 2. This is partially to emphasize that Tags are not the same as a name, and to be consistent with the convention that the most important/least optional values come first.
- Some conventions for input ordering of some priming functions were changed (e.g. instead of previous `prime(ITensor A, IndexType t, int pinc)`, changed to `prime(ITensor A, int pinc, TagSet ts)`, where `prime(ITensor A, TagSet ts) = prime(ITensor A, 1, TagSet ts)` for convenience).
- Some priming functions were removed for simplicity (mostly the more sophisticated ones taking lists of Indices/IndexTypes to prime). We can add some back if they are still needed.
- I changed some of the naming conventions to be more consistent about camel casing (mapprime -> mapPrime, for example), I can add back in the older versions to not break user code and make deprecation smoother if this is the way we want to go.
- In most cases, it is sufficient where an IndexType was used to simply make the change `Link -> "Link"` or `Site -> "Site"` (note that functions are mostly defined to accept a `TagSet` but `const char*` is automatically converted to a `TagSet`, so in general `TagSet`s don't need to be used explicitly). We could even make global definitions of `Link = TagSet("Link")` and `Site = TagSet("Site")` in which case a lot of uses of `IndexType` could just work as is.
- Removed `nameint(std::string s, int n) = format("%s%d",s,n)` convenience function, since Index names have been removed (I think generally using `format()` is easy enough anyway).
- (I also changed the name of the `order()` function to `permute()` because the name `order()` clashes with getting the number of indices of a tensor, and `permute()` is more conventional a name for what it is actually doing, but that can be reverted or done in a separate pull request).

I'm sure there are some more breaking changes but I can update this list as I think of them.

Some things that are still left to do:
- [ ] Add tests for all of the functionality of adding, removing and changing the Tags of Indices, IndexSets and ITensors.
- [ ] Add *Tags functions for MPS and MPO (addTags, removeTags, etc. for site and link indices)
- [ ] Create reading and writing of `TagSet`s so the ITensors can be properly read and written.
- [ ] Choose some tag conventions for different factorizations as well as for the MPS and MPO class, i.e. the right link on site 5 of an `MPS` might have the convention that it has the tags "Link,MPS,5" for example. This could be controlled by the user through the args system.
 - [x] Change `index()` to `findIndexExact()`, add `findIndex()` function


emstoudenmire:
Hi Matt, looks great. One question about the examples: is there a new function called "index" with a lower case "i"? Or is that supposed to be the constructor the Index class?

mtfishman:
It is a function, but I don't like the name of it and I am also not settled on the functionality it should have. The functionality right now is that `index(ITensor A, TagSet ts[, int plev])` searches ITensor `A` for an Index with a TagSet exactly equal to `ts` and a prime level equal to `plev` (where `plev` doesn't have to be specified).

I think a better name should be something like `findIndexExact()`, and have a separate function `findIndex()` that finds an Index that only contains some, but not all of the tags `ts` specified by the user. For concreteness, say we have the following situation:

```c++
auto s = Index(2,"Site,3");
auto l = Index(3,"Link");
auto la = Index(3,"Link,a");
auto lb = Index(3,"Link,b");

ITensor A(s,l,la,lb);

// We want to grab Index `l` from ITensor `A`.
// Here we need an exact match of the tags, since `la` and `lb` also
// have "Link" tags
auto l_found = findIndexExact(A,"Link");  // l_found==l

// We want to grab Index `s`. We know there is only one Index
// with tag "Site" and don't want to specify all of the other tags
auto s_found = findIndex(A,"Site");  // s_found==s

// To grab Index `la`, either function would be fine to use
auto la_found = findIndexExact(A,"Link,a");  // la_found==la
// Or more conveniently:
//auto la_found = findIndex(A,"a");  // la_found==la
```

Another question is how strict these functions should be, i.e. if they should throw errors if multiple matches are found. That could possibly be a check in debug mode.

This is similar to the current `findtype(Itensor A, IndexType T)` function, where an Index of the specified type is found and output, but it becomes both more involved and more powerful using tags. There is also currently a `findindex(ITensor A, Cond cond)`, where a specified condition `cond(i)` for Index `i` in ITensor `A` needs to be matched, that sort of thing might still be useful if people want more sophisticated logic for finding an Index.

itensor/tagset.h
```diff
@@ -0,0 +1,228 @@
...
+class TagSet
+    {
+    public:
+    using tags_type = std::array<Tag,MAX_TAGS>;
+    private:
+    tags_type tags_;
+    size_t size_ = 0;
+    public:
+
+    TagSet() {};
```
emstoudenmire:
```suggestion
    TagSet() {}
```
^ Shouldn't have a semicolon after a method declaration

itensor/tagset.h
```diff
@@ -0,0 +1,228 @@
...
+
+    //
+    // Convert a character array to a TagSet.
+    // Parses the array for ',' and splits up into tags, for example:
+    // "a,b,c" -> Tag("a"),Tag("b"),Tag("c")
+    TagSet(const char* t);
+
+    std::string
+    to_string()
```
emstoudenmire: Usually I've been naming class methods using "camel case" so I'd use toString instead of to_string. Was the choice of to_string to match another existing convention though? I think using c_str below is good, in contrast, because it is parallel to the similar method that std::string defines. (So ok to break the rules sometimes, especially for advanced/developer functionality.)

itensor/tagset.h
```diff
@@ -0,0 +1,228 @@
...
+    };
+
+inline size_t
+size(TagSet const& t)
+    {
+    return t.size();
+    }
+
+inline bool
+operator==(TagSet const& t1, TagSet const& t2)
```
emstoudenmire: Some compilers will complain that not all paths through this function will return a bool (even though logically it's not true). So a better design I think would be to remove the else branch and just put the for loop in the main function body. Then the first if will either return or fail, calling the "else" code anyway which is guaranteed to return.

itensor/tagset.h
```diff
@@ -0,0 +1,228 @@
...
+
+// Adds a Tag to a TagSet. Does nothing if the tag already exists,
+// otherwise places the tag in a specified ordering
+inline void TagSet::
+addTag(const Tag& t)
+    {
+    if(size_ == MAX_TAGS) error("Too many tags already, cannot add more. If you want more, consider raising MAX_TAGS.");
+    if(this->hasTag(t) == -1 && t != Tag())  // If Tag is not found and is not empty, add it
+        {
+        size_t i;
```
emstoudenmire: Better to go ahead and initialize i right here, then just omit the initialization inside the for loop
```suggestion
        auto i = size_;
        for(; i > 0; --i)
```

itensor/tagset.h
```diff
@@ -0,0 +1,228 @@
...
+            }
+        tags_[i] = t;
+        size_++;
+        }
+    }
+
+inline void TagSet::
+addTags(const TagSet& ts)
+    {
+    for(size_t i = 0; i<ts.size(); ++i)
```
emstoudenmire: For many of these loops, you can use the "range" feature we have in ITensor. It has the advantage that it lets you use auto to "stick" the type of the loop counter so that if we ever change the type all the code will just automatically remain correct.
```suggestion
    for(auto i : range(ts.size()))
```

emstoudenmire:
Yes I think findIndexExact is a better name than anything else I could think of right now. I briefly toyed with the idea of extending the string notation, something like:
```
findIndex(T,"Site,...");
```
for matching a "Site" tag but being ok if there are other tags too. Whereas just "Site" would be like the exact case.

But I think the above would likely be too confusing and too inconvenient, because the more casual findIndex behavior is what I suspect people will use more often. Plus it would be annoying if one used findIndex(T,"Site") one day and it worked because the index only had one tag, then later one adds more tags and it breaks. Then you'd have to change your code to say findIndex(T,"Site,...") or always be careful to put this "just in case". So I don't like it.

Therefore I think have two different functions is clearer.

itensor/tagset.h
mtfishman: There was no particular reason for the name, I agree we should be consistent about using camel-casing.

itensor/tagset.h
mtfishman: I'll change to using `range` whenever I can, it is definitely not ideal to have to specify `size_t` vs. `int` depending on what limit is being used.

mtfishman:
I pushed changes based on your comments (along with some other code cleanup).

krvajal:
There seems to be some merge conflicts. You should merge the latest changes from master into your branch and solve the conflicts before expecting your PR to be merged

krvajal:
Then I think this can be closed to avoid clutter and confusion.

mtfishman:
Closing, as this work was moved to the rc3 branch.

PR #188, opened 2018-11-20
New function errorMPOProd

ryanlevy:
As discussed in #185, here is a more complete implementation of the MPO error function. Note that the `abs` in line 573 is necessary due to underflow when normalization is large.


emstoudenmire:
This looks great! Thanks Ryan. I added your name to the About part of the ITensor website too.

PR #189, opened 2018-11-25
New random MPO constructor

ryanlevy:
I would like to propose the following interface to create random MPOs directly rather than the AutoMPO method. 
```c++
/* create an MPO with a random tensor on every site with uniform dimension k on Link bonds
 * optional argument "Orthogonalize":true in order to 
 * automatically orthogonalize after creation
 * and "Scale" which does Aref(j)*=scale
 */
int k = 2;
auto H1 = MPO(sites,k);

auto OrthoH = MPO(sites,k,{"Orthogonalize",true}); 

/* create an MPO with a random tensor on every site 
 * with each Link bond dimension specified by a vector
 */
std::vector<int> bondList = {2,3,5,3,2};
auto H2 = MPO(sites,bondList);
//can create directly in constructor
auto H3 = MPO(sites,{2,2,2,2,2},{"Orthogonalize",true,"Scale",0.2});

```
I have only implemented a non-QN version as I'm not sure what a QN one should look like. If anyone has comments/suggestions I'd be happy to implement it!

### DIFF
--- a/itensor/mps/mpo.cc
+++ b/itensor/mps/mpo.cc
@@ -64,6 +64,50 @@ template
 MPOt<IQTensor>::
 MPOt(const SiteSet& sites, Real _logrefNorm);
 
+template<class Tensor>
+MPOt<Tensor>::
+MPOt(const SiteSet& sites, 
+     std::vector<int> const& bondList,
+     const Args& args) 
+    : 
+    Parent(sites)
+    { 
+    bool doOrthogonalize = args.getBool("Orthogonalize",false);
+    fillRandomMPO(*this,sites,bondList,args);
+
+    if (doOrthogonalize) orthogonalize(args);
+    }
+template
+MPOt<ITensor>::
+MPOt(const SiteSet& sites, 
+     std::vector<int> const& bondList,
+     const Args& args);
+
+template<class Tensor>
+MPOt<Tensor>::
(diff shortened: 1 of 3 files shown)


emstoudenmire:
Hi Ryan,
Thanks for this suggestion and pull request. Matt Fishman and I were discussing it, and we are going to hold off on making it one of the MPO constructors right away. However, it could be very useful to have a function that generates random MPOs, at least for the case of ITensors (dense tensors). The case of IQMPO's is rather more tricky. 

It does raise the question of what is the "right" or really "most useful" distribution over random MPOs. Probably just one where the elements are individually Gaussian distributed is already very useful for many purposes, such as unit testing. 

For the case of random MPS's, I proposed that maybe what is really wanted is a random unitary circuit of a certain depth applied to a product state. So we may add a function randomMPS for that.

ryanlevy:
Alright, I can just move this to a personal function for my code then. I would still like to advocate for a better method of creating MPOs than using AutoMPO then overwriting everything though.

And I +1 the randomMPS idea, I think that would be very useful

emstoudenmire:
Thanks a lot for the feedback Ryan & thanks for understanding about this decision for now. 

If you wouldn't mind reconstituting your code as a stand-alone function called randomMPO -> MPO, I would merge that pull request.

PR #190, opened 2018-12-07
Add link to Install file

krvajal:
Added a link to the Install instruction so it can be looked up easily.


emstoudenmire:
Thanks!

PR #191, opened 2018-12-11
combiner constructor accepts IndexSet/IQIndexSet

mtfishman:
As suggested by @kshyatt.

### DIFF
--- a/itensor/iqtensor.cc
+++ b/itensor/iqtensor.cc
@@ -351,10 +351,10 @@ div(IQTensor const& T)
     }
 
 IQTensor
-combiner(std::vector<IQIndex> cinds,
+combiner(IQIndexSet const& inds,
          Args const& args)
     {
-    if(cinds.empty()) Error("No indices passed to combiner");
+    if(inds.empty()) Error("No indices passed to combiner");
     auto cname = args.getString("IndexName","cmb");
     auto itype = getIndexType(args,"IndexType",Link);
 
@@ -368,7 +368,7 @@ combiner(std::vector<IQIndex> cinds,
         //If not specified by user, make combined IQIndex
         //point Out unless all combined indices have In arrows.
         auto allin = true;
-        for(auto& i : cinds) 
+        for(auto& i : inds) 
             if(i.dir() != In)
                 {
                 allin = false;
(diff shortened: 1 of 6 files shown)


emstoudenmire:
This is a good feature to have, but can it also accept a std::vector of indices? I guess it can if one converts the std::vector to an IndexSet or IQIndexSet?

The reason is that in some user code built on ITensor, it's convenient to add indices one-by-one to a std::vector using push_back, then make a combiner out of that set.

emstoudenmire:
Agreed. But I do believe this patch does not keep the old functionality, but instead replaces the std::vector argument with an IndexSet argument. Maybe the automatic conversion of vector<Index> to IndexSet kicks in though, and this function can accept either one? Or maybe by changing the argument from const-ref to pass-by-value it will accept both?

mtfishman:
With this pull request I added `combiner([IQ]IndexSet)`, and now `combiner(vector<[IQ]Index> inds) = combiner([IQ]IndexSet(inds))`.

As a side issue, I was thinking it might be nice to have a function to add an Index to an IndexSet, like `addIndex(IndexSet,Index) -> IndexSet`.

mtfishman:
I explicitly added the function:

```c++
IQTensor
combiner(std::vector<IQIndex> const& inds,
         Args const& args)
    {
    return combiner(IQIndexSet(inds),args);
    }
```

mtfishman:
No worries! In `combiner(IndexSet)` it basically implements `addIndex(IndexSet,Index)` so it will be easy to add. I suppose we could even have an `addIndices(IndexSet,IndexSet) -> IndexSet` function that adds one IndexSet to another (though I am not sure what the name should be or if there is a good use case).

emstoudenmire:
Yes, I would say let's be conservative about adding things like addIndices. Although it could certainly have a use, if we can't find anywhere it is being used then it's better to wait for such a case to arise.

But .addIndex for a single index could definitely replace some spots in the code where a std::vector is first filled with indices and then used to make an IndexSet. (I think there are a few such places...could be wrong. There is also a helper class called "RangeBuilder" for this purpose but maybe it can be deprecated and just using .addIndex is simpler and easier.)

mtfishman:
That sounds good. It would be nice to consistently use IndexSets if possible.

PR #195, opened 2019-01-03
Add IgnoreDegeneracy flag to svd, diagHermitian, and dmrg

mtfishman:
This adds an argument "IgnoreDegeneracy" to the functions svd(), diagHermitian(), and dmrg() which allows a user to specify whether or not degeneracies in the singular value spectrum or eigenvalue spectrum are ignored when truncating.

In this pull request, "IgnoreDegeneracy" defaults to true for svd() and diagHermitian(), and defaults to false for dmrg().

I am definitely open to suggestions about the name of the flag and the defaults. With these defaults, it makes it so that simple situations like:
```c++
auto i = IQIndex("i",Index("i",2),QN(0));
auto A = IQTensor(i,dag(prime(i)));
A.set(1,1,1.0);
A.set(2,2,1.0);

IQTensor U(i),S,V;
svd(A,U,S,V,{"Maxm=",1});
```
"just work" now, i.e. truncates one of the singular values and outputs potentially random singular vectors. This case currently throws a mysterious error since it truncates the entire set of degenerate singular values (this pull request also improves the debug error message if {"IgnoreDegeneracy=",false} is added to the above args list).

### DIFF
--- a/itensor/decomp.cc
+++ b/itensor/decomp.cc
@@ -114,6 +114,7 @@ truncate(Vector & P,
          bool doRelCutoff,
          Args const& args)
     {
+    auto ignore_degeneracy = args.getBool("IgnoreDegeneracy",true);
     long origm = P.size();
     long n = origm-1;
     Real docut = 0;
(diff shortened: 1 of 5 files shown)


emstoudenmire:
Thanks for looking into this so closely; it's a really good improvement I think. Even though I went ahead and pulled it, I do think a better name could be something like "KeepDegenerateEigs" or similar (especially a shorter name would be welcome). The reason is that it's easier for me to understand what "keeping degenerate eigenvalues" does since it describes a situation and the action taken, whereas ignore degeneracy sort of leaves it unclear to me what happens exactly once a degeneracy occurs. 

Choosing the right default seems harder. I think the previous behavior was probably a bit to DMRG-oriented. A better default would be the one that gives the most sensible behavior in the most general cases and avoids weird cases like the one you identified with Maxm=1. That's definitely a very buggy behavior!

PR #196, opened 2019-01-23
added the option for the user to set the index type for IQTensors in …

jansendavid:
…diagHImpl<IQTensor...>, which already worked for normal  ITensors


emstoudenmire:
Great! Thank you

PR #200, opened 2019-02-06
Tag convention for MPS and MPO, with tests

mtfishman:
I have tried to make the tag conventions a bit more sane for MPS and MPO.

Both now have the conventions that the sites have tags "Site,n=#", and links have conventions "Link,l=#". I added tests to check that some basic MPS and MPO operations like gauging, adding, applyMPO, etc respect the tag convention. 

I think I have missed some spots, and it would be nice to have a good way for users to specify their own convention, but I think this is a good start.

### DIFF
--- a/itensor/decomp.h
+++ b/itensor/decomp.h
@@ -303,7 +303,8 @@ denmatDecomp(ITensor const& AA,
     //Apply combiner
     START_TIMER(8)
     //TODO: decide on a tag convention for denmatDecomp
-    auto itagset = getTagSet(args,"Tags","Link,MID");
+    auto itagset = getTagSet(args,"Tags","Link");
+    args.add("Tags",toString(itagset));
     auto cmb = combiner(std::move(cinds),{"Tags",toString(itagset)});
     auto ci = cmb.inds().front();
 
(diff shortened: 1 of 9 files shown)


mtfishman:
(Also, I fixed a warning about comparing unsigned and signed integers from some of the new QN code.)

emstoudenmire:
Great! Thanks for the hard work on this. I think it's a big improvement.

emstoudenmire:
Just tried out the DMRG sample code after this change and it's great to print out the MPO and MPS and see the helpful tags still in there. It makes it so much easier to see by eye which indices match.

mtfishman:
I was wondering, maybe it would be good to have a different tag convention for the link of an MPS and MPO? For example, if I am making a Hamiltonian environment:
```c++
auto E1 = psi.A(1)*H.A(1)*prime(dag(psi.A(1)),"Link");
```
It might be nice to be able to distinguish the link index of the MPO vs. the MPS (both for printing and for Index manipulations). Maybe something like "hl=1" for MPO links and "l=1" for MPS links?

emstoudenmire:
Hi Matt, perhaps - let’s think about it, but it may be always ultimately hard to get this absolutely perfect, given the ‘dynamic’ setting (i.e. SVD always stamping out new indices) and that users might make their own MPOs and MPS etc.

So I’d be supportive of, say, naming the H links “hl=7” etc. but of course we can’t rely on that naming to be there in any library algorithms since users could input their own custom MPOs.

PR #202, opened 2019-02-20
Add back Arnoldi with tests

mtfishman:
Here is the Arnoldi method you sent me a while ago, updated a bit and included with the other iterative solvers. I also included some simple unit tests, it looks like it works well!


emstoudenmire:
Very welcome add! This is a key thing to have back in ITensor

emstoudenmire:
And great that it has tests too

PR #211, opened 2019-03-12
Silent DMRG

1ucasvb:
- Fixed a small bug, in which the "verbose" option on DMRG worked for IQMPOs but not for MPOs.
- Added an extra setting, "Silent", which implies Quiet and also suppresses all other outputs. (This is desirable if DMRG is used as a sub-routine on an algorithm.)


emstoudenmire:
Yes, I agree this is needed! Thanks -

PR #220, opened 2019-03-29
suggested minor improvements to indextype

jansendavid:
using the stl algorithm instead of a for loop, arguably improves readability and produced significantly less assembly for gcc 8.3.0 and clang 8.0.0 with -O3


emstoudenmire:
Thanks

PR #224, opened 2019-04-05
Minimize Clang Warning

titaschanda:
Clang 8 throws following warning -->
warning: local variable 'B' will be copied despite being returned by name [-Wreturn-std-move]
....
 note: call 'std::move' explicitly to avoid copying

### DIFF
--- a/itensor/itensor_interface_impl.h
+++ b/itensor/itensor_interface_impl.h
@@ -734,7 +734,7 @@ ITensorT<I>
 operator*(ITensorT<I> A, ITensorT<I> const& B) { A *= B; return A; }
 template<typename I>
 ITensorT<I>
-operator*(ITensorT<I> const& A, ITensorT<I>&& B) { B *= A; return B; }
+operator*(ITensorT<I> const& A, ITensorT<I>&& B) { B *= A; return std::move(B); }
 template<typename I>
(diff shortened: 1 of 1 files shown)


emstoudenmire:
Hi Titas,
Thanks! Just out of curiousity, does replacing `std::move(B)` with `std::forward<ITensorT<I>>(B)` work and also get rid of the warning? That might be a better solution, because std::forward adapts to the case of whether B is a temporary (an r-value) or not.

titaschanda:
Hi Miles, yes, std::forward should also stop the warning, but I believe that std::forward is not necessary here, as the warning comes only when B is taken as rvalue reference (i.e., in function definitions). Where B is passed as const lvalue reference, no warning is generated, as it will call the const& overloads. As the function templates in question are not "universal reference", these functions will only be called for rvalues, and std::move is sufficient. Please correct me if I am wrong.

emstoudenmire:
Hi Titas,
Yes, so I was worried about the following: that it would remove the warning in some cases, but then there would be cases where B was *not* an rvalue and then we'd see a different warning about how std::move wasn't needed.

But you're right to point out that this is not a universal reference, always an rvalue reference. So let's definitely add std::move then. Thanks for catching this!

Issue #219, opened 2019-03-25
Concatenate Local Operators


dkweiss31:
This is something user benoit mentioned already on the website under "AutoMPO Syntax: Concatenate local terms" but miles commented that it would be best to file a feature request. I'm wondering if it would be possible to have a function that would allow you to construct an operator involving operators from an arbitrary number of sites, without explicitly needing to write
ampo += "Sz", j, "Sz", j+1, "S+", j+2, ...
I have a system where the number of operators increases with the system size, so for example for a system of size 7 I will have an operator involving all 7 sites (they will have small coefficients out front compared to more local operators so that DMRG should hopefully still be useful). See benoit's post for an example of the pseudocode that would be useful to have. Thanks!

PR #229, opened 2019-04-20
Add AutoMPO concat function for local operators

ryanlevy:
Addressing #219 and [this forum post](http://itensor.org/support/1212/autompo-syntax-concatenate-local-terms?show=1212), this provides a concatenation function for local operators and `AutoMPO`

Usage:
```c++
auto ampo = AutoMPO(sites);

//loop in a function
for(auto n: range1(N))
   ampo.concat("Sz",n);
ampo.combine(); //done concating, does the += step

//For coefficients and a function
Real J = -1.7;
//same as ampo += J,"Sz",1,"Sz",2,"Sz",3;
ampo.concat(J);
for(auto n: range1(3))
    ampo.concat("Sz",n);
ampo.combine();

//don't like the comma operator? This works too
//same as ampo += 3.14,"Nup",2,"Ndn",4,"Nupdn",7";
ampo.concat(3.14,"Nup",2)
    .concat("Ndn",4,"Nupdn",7) //takes any number of arguments
    .combine();
``` 
Caveat: only one concat sequence is allowed at a time. 

I'm not 100% in love with the names, so please feel free to suggest alternatives.

### DIFF
--- a/itensor/mps/autompo.h
+++ b/itensor/mps/autompo.h
@@ -190,10 +190,11 @@ class AutoMPO
 
     public:
 
-    AutoMPO() { }
+    AutoMPO(): concatAccumulator(nullptr) 
+        { }
 
     AutoMPO(SiteSet const& sites) 
-      : sites_(sites)
+      : sites_(sites), concatAccumulator(nullptr)
         { }
 
     SiteSet const&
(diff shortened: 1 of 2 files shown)


dkweiss31:
Thank you!

PR #240, opened 2019-05-13
Changed project_template's Makefile's LIBRARY_DIR

ajrheng:
Changed project_template's Makefile's LIBRARY_DIR to (HOME)/itensor instead of (HOME)/software/itensor so it is consistent with installation instructions on the website, where one would normally clone ITensor to the (HOME) folder directly (on a Unix system), rather than have an intermediate /software/ folder. Without this change the Makefile in the project_template does not work out of the box (errors that folder does not exist), and might confuse some people.


emstoudenmire:
Hi, this is a good suggestion so I'll pull the change. Thanks for making ITensor better!

PR #244, opened 2019-06-07
added possibility of defining Emp to the bosonic siteset

jansendavid:
This allows for a mixed site set like the Holstein MixedSiteSet<Fermion, Boson> to initialize a complete state as unoccupied with setAll("Emp")


emstoudenmire:
Good idea, thanks!

PR #245, opened 2019-07-01
allowed for the user to the the accuary

jansendavid:
allowed for the user to the the accuary of the approcimation, deafult is still 100


emstoudenmire:
Hi, did you have a case where you needed this to be a different size? If so, was it larger or smaller. Thanks!

jansendavid:
Yes,
I had some cases where I could reproduce the same results for a certain accuracy with a smaller size

PR #252, opened 2019-08-13
Fix eigen

mtfishman:
Eigen had the following bug:

```C++
  auto i = Index(2,"i"),
       j = Index(2,"j"),
       k = Index(2,"k");
  auto I = IndexSet(i,j,k);
  auto Ip = prime(I);

  auto A = randomITensor({I,Ip});

  auto [V,E] = eigen(A,{"Tags=","test"});

  PrintData(norm(A*V - prime(V)*E));  // Prints non-zero, should be zero
```

If we replace `auto A = randomITensor({Ip,I})`, the `eigen` function works. The problem seems pretty subtle, in that the function forms a MatrixRef out of the data of the ITensor for indices of the form `i',j',k',...,i,j,k,...`. If it gets an ITensor with the indices above, it makes a transposed MatrixRef, and then calls the internal `eigen` for matrices, which uses a special version if the MatrixRef is transposed. It seems like the output of `eigen(MatrixRef,...)` is not being used correctly in that case (it seems to end up with something related to the left eigenvectors, while `eigen(ITensor,...)` is meant to output the right eigenvectors).

The fix I used was just to use `permute(A,{prime(c),c})` where `c` is the combined index of `{i,j,k,...}` so the MatrixRef never needs to be transposed. This may lead to some extra unnecessary copying of data, but I think the time that takes is much less then the time to call `eigen` anyway. There should be a deeper way to fix the bug, but I tried to debug it for a while and couldn't figure out how to get whatever was being output by `eigen` in the transposed case into the correct set of eigenvectors.

### DIFF
--- a/itensor/decomp.cc
+++ b/itensor/decomp.cc
@@ -543,13 +543,15 @@ factor(ITensor const& T,
 
 //TODO: create a tag convention
 template<typename value_type>
-void 
+Index 
 eigDecompImpl(ITensor T, 
               ITensor & L, 
               ITensor & R, 
               ITensor & D,
               Args const& args)
     {
+    // New index listing eigenvectors
+    Index newmid;
     if(not hasQNs(T))
         {
         auto full = args.getBool("FullDecomp",false);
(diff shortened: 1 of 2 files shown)


emstoudenmire:
Wow, that sounds subtle and frustrating. Thanks for finding & fixing this.

PR #259, opened 2019-08-28
Add .removeQNs() function for SiteSet

mtfishman:
This is one solution to help people create operators that don't conserve QNs if the SiteSet they initially created conserves QNs.

### DIFF
--- a/itensor/mps/siteset.h
+++ b/itensor/mps/siteset.h
@@ -61,6 +61,9 @@ class SiteSet
     int 
     length() const;
 
+    SiteSet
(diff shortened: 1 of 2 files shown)


mtfishman:
I realized that this isn't general enough of a solution, it works for SiteSets that inherit from BasicSiteSet or MixedSiteSet but not from ones (like `SpinOne`) that don't. I think a more involved approach is needed to implement this feature so that every custom SiteSet that is defined doesn't require having to define their own `removeQNs` function.

PR #263, opened 2019-09-10
Added TDVP routine

mingruyang:
[1] Added one-site and two-site TDVP algorithm
[2] Added a Krylov subspace method to calculate exp(-tH)*v and exp(-itH)*v
[3] Added the irreducible Pade approximation to do the exponential of a general small dense matrix
[4] Added wrapper to calculate the norm of a matrix and do the LU factorization
[5] Added support of 0 and 1 center sites for localmpo and localop
[6] Corrected a bug in shift() of localmpo.h
[7] Added a sample code to use the tdvp function

To be done:
[1] Chebyshev approximation (the Chebyshev method can fail when solving some specific matrix, so I removed that implementation when I pushed)
[2] Use template<typename T> for expPade
[3] Add an observer for TDVP
[4] Add QR decomposition to replace the SVD in one-site algorithm

### DIFF
--- a/itensor/all_mps.h
+++ b/itensor/all_mps.h
@@ -35,6 +35,7 @@
 #include "itensor/util/stats.h"
 
 #include "itensor/mps/dmrg.h"
+#include "itensor/mps/tdvp.h"
 #include "itensor/mps/tevol.h"
 #include "itensor/mps/autompo.h"
 
--- a/itensor/iterativesolvers.h
+++ b/itensor/iterativesolvers.h
@@ -62,6 +62,22 @@ gmres(BigMatrixT const& A,
       Args const& args = Args::global());
 
 //
+// Use the Krylov subspace method to compute
+// phi' = exp(-t*A)*phi
+// or
+// phi' = exp(-it*A)*phi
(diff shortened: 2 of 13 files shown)


mtfishman:
Hi Mingru,

Sorry for being slow about responding to this. Over all it looks good, though I need to look it over more carefully. Thanks for the contribution!

Since it is a very large pull request, I think it would be easier for the review to split this up into a few parts, as follows:

1. New linear algebra routines (the new krylov method, Pade approximation, and LAPACK wrappers)
2. Extension of the LocalMPO/LocalOp for other unit cell sizes
3. TDVP

Could you submit separate pull requests for 1. and 2.? 

Also, I think for now we want to have the TDVP routine in a separate repository (like iDMRG: https://github.com/ITensor/iDMRG). This is because we want to work out some of the best practices for how to use it before giving it to people as a "black box" algorithm. For example, I believe @alexwie found that the best way to increase the bond dimension at the start of the algorithm is to use a few steps of TEBD, so this is something that we should explore more and provide a nice interface for.

Cheers,
Matt

PR #265, opened 2019-09-20
Changed inheritance of MPO private->protected

brendenroberts:
So that subclasses of MPO can access members.


emstoudenmire (approved):

emstoudenmire:
Thanks Brenden. I'm good with this change - if you are personally considering inheriting from MPO, though, I'd recommend also considering some other patterns, such as making generic (template) functions which can work on your new type as well as MPO, or making MPO a data member of your new type (composition).

PR #272, opened 2019-10-02
Allowed localmpo/localmposet/localop to have 0 or 1 or 2 center sites.

mingruyang:
Unittest added.

### DIFF
--- a/itensor/decomp.cc
+++ b/itensor/decomp.cc
@@ -270,7 +270,8 @@ svd(ITensor const& AA, IndexSet const& Uis,
 
 std::tuple<ITensor,ITensor>
 polar(ITensor const& T,
-      IndexSet const& Uis)
+      IndexSet const& Uis,
+      Args const& args)
     {
     auto [U,S,V] = svd(T,Uis);
     auto u = commonIndex(S,U);
@@ -279,6 +280,26 @@ polar(ITensor const& T,
     auto Vp = prime(V,Vis)*delta(v,u);
     auto Q = U*Vp;
     auto P = dag(Vp)*S*V;
+    auto qis = commonInds(Q,P);
+
+    // Add tags to the internal indices
+    auto ts = getTagSet(args,"AddTags","");
+
+    // Prime the internal indices.
+    // Prime by one less than specified,
+    // since they are already primed by one
+    // above.
+    auto plinc = args.getInt("Prime",1)-1;
+
+    if(primeLevel(ts) >= 0) Error("In polar, specify a prime level increment with the Prime argument");
+    if(ts=="" && plinc==-1) Error("In polar, must either increment prime level or add tags to new indices");
+
+    Q.addTags(ts,qis);
+    P.addTags(ts,qis);
+    qis.addTags(ts);
+    Q.prime(plinc,qis);
+    P.prime(plinc,qis);
+
     return std::tuple<ITensor,ITensor>(Q,P);
     }
 
--- a/itensor/decomp.h
+++ b/itensor/decomp.h
@@ -47,7 +47,8 @@ svd(ITensor const& AA, Index const& i1, IndsArgs&&... indsargs);
 
 std::tuple<ITensor,ITensor>
 polar(ITensor const& T,
-      IndexSet const& Uis);
+      IndexSet const& Uis,
+      Args const& args = Args::global());
 
 //
 // The "factor" decomposition is based on the SVD,
--- a/itensor/mps/localmpo.h
+++ b/itensor/mps/localmpo.h
@@ -65,6 +65,14 @@ class LocalMPO
     //
     LocalMPO(MPS const& Psi, 
              Args const& args = Args::global());
+    
+    //
+    //Use left and right boundary tensors LH and RH. Ok if both boundary 
+    //tensors are not default-constructed.
+    //
+    LocalMPO(const ITensor& LH, 
+             const ITensor& RH,
+             const Args& args = Args::global());
 
     //
     //Use an MPO having boundary indices capped off by left and
@@ -177,8 +185,9 @@ class LocalMPO
     void
     numCenter(int val) 
         { 
-        if(val < 1) Error("numCenter must be set >= 1");
+        if(val < 0 || val > 2) Error("numCenter must be set 0 or 1 or 2");
         nc_ = val; 
+		lop_.numCenter(val);
         }
 
     size_t
@@ -292,6 +301,21 @@ LocalMPO(const MPS& Psi,
     }
 
 inline LocalMPO::
+LocalMPO(const ITensor& LH, const ITensor& RH,
+         const Args& args)
+    : Op_(0),
+      PH_(2),
+      LHlim_(0),
+      RHlim_(1),
+      nc_(0),
+      Psi_(0)
+    { 
+    PH_[0] = LH;
+    PH_[1] = RH;
+    lop_.update(L(), R());//nc_ must be set to 0 in this case
+    }
+
+inline LocalMPO::
 LocalMPO(const MPO& H, 
          const ITensor& LH, const ITensor& RH,
          const Args& args)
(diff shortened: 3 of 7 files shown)


mingruyang:
Previously I got *this point to NULL if I set Op1_ = nullptr in the update  function in LocalOp class. But this bug disappears this time. I'm not sure what happened.

mtfishman:
Thanks for the new pull request, I am taking a look at it now. The LocalOp part looks good so far.

An immediate question I have is, how do you think this design would work for an arbitrary number of center sites? I know that you have dealt with larger numbers of center sites for DMRG (like 3 or more, right?). I don't think we need to address the general case in this pull request, but it is good to make sure that the interface would generalize in case we want to add that in the future.

mtfishman:
To answer my own question a bit, it looks like at a minimum we would need `LocalOp` to store `vector<ITensor>` of length `nc_` instead of a fixed number of "local operators", right now `Op1_` and `Op2_`. In theory, it seems like the interface you made is fairly general, so that could be an internal change and then a `LocalOp` could be made with an arbitrary number of operators (or a vector of operators).

mtfishman:
Over all, testing `LocalOp` a bit, that interface looks quite good.

I am having a bit of trouble using `LocalMPO` when `.numCenter() == 0`, however. Here is a code I tried:

```C++
  int N = 10;
  auto sites = SpinHalf(N);
  auto ampo = AutoMPO(sites);
  for(int j = 1; j < N; ++j)
      {
      ampo += 0.5,"S+",j,"S-",j+1;
      ampo += 0.5,"S-",j,"S+",j+1;
      ampo +=     "Sz",j,"Sz",j+1;
      }
  auto H = toMPO(ampo);
  auto psi = MPS(InitState(sites,"Up"));

  auto b = 4;

  auto Hpsi = LocalMPO(H,{"NumCenter=",0});

  PrintData(Hpsi.numCenter());

  psi.position(b-1);

  // We need to split off the 0-site wavefunction,
  // we can do that with a polar decomposition

  auto l = linkIndex(psi,b-2);
  auto s = siteIndex(psi,b-1);
  auto [U,C] = polar(psi(b-1),{l,s});

  // By default, the new Index is the prime of the
  // right indices. However, this could cause clashes
  // in .product(...) since the ket link indices are also
  // primed. To avoid this, I am changing the tags
  // (polar should have an interface for this but doesn't
  // right now).
  auto u = commonIndex(U,C);
  auto newtag = "Link,left,l="+str(b-1);
  U.setTags(newtag,u);
  C.setTags(newtag,u);

  // We update the MPS with the U tensor, so the
  // effective Hamiltonian is made correctly
  psi.ref(b-1) = U;

  Hpsi.position(b,psi);

  auto phi = C;
  auto Hphi = phi;

  // This is what we want product to calculate:
  PrintData(noPrime(phi*Hpsi.L()*Hpsi.R()));

  // This doesn't work:
  Hpsi.product(phi,Hphi);

  PrintData(norm(Hphi-noPrime(phi*Hpsi.L()*Hpsi.R())));
```

The `.product(...)` function errors with "LocalOp is null", so it looks like `.product` doesn't handle the 0-site case right now. I just wanted to check with you to see if this is something you are still working on, or if I am using it incorrectly. Was this the kind of way you were imagining this function to be used in the 0-site case?

mingruyang:
Hi Matt,

.product() in LocalOp does handle the 0 center site case. Since when we construct the LocalOp, the number of center sites can only be 0, 1 or 2. So if nc_ is not 2 or 1, it is assumed to have nc_ = 0 and only the L() or R() will be producted.

I think this bug is exactly what I mentioned to you in my pull request. 

If we have 0 center sites, .update() will set Op1_  a nullptr. *this will be null after that. So you got the error message “LocalOp is null”.
Originally I didn’t allow user to set 0 center sites when construct the object. One can get 0 center sites later by set .numCenter() and .position() in LocalMPO. And in the .update(L(),R()) function of LocalOp, I commented out the lines “Op_1 = nullptr; Op_2 = nullptr”. So *this will not be null this way. But it is absolutely ugly.

I think one way to deal with the problem is that we use std::vector in localOp to allow mutable center sites, just like what you said.

But I’m not sure if there is another easier way to deal with it if we want to maintain the current settings.

Best,
Mingru

mtfishman:
I see, it is definitely a subtle question of how to add this functionality in a nice way and not make breaking changes.

Actually I realized I was testing the `LocalOp` functionality incorrectly before. I thought the number of center sites was determined by the number of inputs to the constructor (instead of set by the args), so I see the same problem there as well:

```C++
auto l0 = Index(2,"Link");
auto r0 = Index(2,"Link");
auto h0 = Index(2,"Link");

auto L0 = randomITensor(l0,prime(l0),h0);
auto R0 = randomITensor(r0,prime(r0),h0);

auto psi0 = randomITensor(l0,r0);
auto H0 = LocalOp(L0,R0,{"NumCenter=",0});

auto Hpsi0 = psi0;
H0.product(psi0,Hpsi0); // Throws the same error
```

I guess the problem comes down to the definition of boolean evaluation of a `LocalOp` here:

https://github.com/ITensor/ITensor/blob/krylov/itensor/mps/localop.h#L162

which is what triggers the error here:

https://github.com/ITensor/ITensor/blob/krylov/itensor/mps/localop.h#L321

in the `.product(...)` function. Perhaps we can just extend the definition of boolean evaluation of a `LocalOp` to check the `numCenter()` and use a different definition depending on the number of center sites, for example:

```C++
explicit operator bool() const
  {
  if(numCenter()==0) { return !LIsNull() || !RisNull(); }
  else if(numCenter()==1) ...
  }
```

since we don't care if `Op1_` and `Op2_` are null for the 0-site case.

mtfishman:
I also have another suggestion for the design. It may be better if _all_ of the `LocalOp` constructors checked for the "NumCenter" argument, like the one accepting two ITensors does (https://github.com/ITensor/ITensor/blob/krylov/itensor/mps/localop.h#L199). Then, errors can be thrown (or null values set) if the number of ITensors input doesn't match with the number expected for the specified number of center sites. All of the functions can default to a 2-site center if no "NumCenter" arg is input, which would guarantee backwards compatibility.

An advantage of this is that the current design is a bit confusing, since some constructors determine the `numCenter()` from the number of input ITensors, and some determine it from the "NumCenter" argument. I think actually the "NumCenter" argument is the better way to go, since the number of inputs can be ambiguous.

mingruyang:
Hi Matt,
I changed the definition of the boolean evaluation. Could you please check if the same error still appears?

mtfishman:
The changes you made look good and fixed the errors I was seeing. I cleaned up the code a little bit and added more unit tests for the `LocalMPO` and `LocalMPOSet` product function.

I also noticed that the "NumCenter" arg wasn't being used in the `LocalMPOSet` and `LocalMPO_MPS` constructors so I fixed that.

Do you think there is anything else to add to this pull request? Does it cover everything you need for TDVP? If so, I can merge it.

Our current plan would still be to put the TDVP code itself in a separate repository (at least for now). Would you be ok with the repository being in the ITensor organization (https://github.com/ITensor)? We will of course give you credit on the website for the contribution! I can help set it up like the iDMRG repository.

mingruyang:
Hi Matt,

Thanks for the improvement!

I think currently that's all I need to do TDVP.

There are several things we can improve in the future:
1. Improve the interface for the observer for TDVP. Currently I just use the DMRG observer. I'm not sure if the TEvolObserver is enough or not.
2. In the expApply, to estimate the error we need a function to evaluate the norm of the LocalMPO, currently I just use a overlap instead. I also tried to just multiply the norm of different part of the LocalMPO. Since I didn't find a efficient way to actually do calculate the norm of the whole LocalMPO. Currently these two alternative ways look fine. But I worried there might be possibilities that they don't work.  
3. Add a QR decomposition routine so the one-site algorithm can be done more efficiently.
4. Add Chebyshev approximation.

I would like the TDVP repository to be in the ITensor organization. Please help me to set it up. Thanks a lot!

mtfishman:
Ok great, those are things we can work on in the future but it is good to have a starting place that works. I will set up the TDVP repository, and then you can push the latest TDVP to that.

mtfishman:
I set up the repository for TDVP here: https://github.com/ITensor/TDVP

You can just push the `tdvp.h` file to that repository, and then we can add some sample codes and tests.

mingruyang:
I have made the tdvp.h and a sample code. But my push is denied because I don't have the permission.

mtfishman:
Sorry about that, I just added you as a collaborator. Could you try again?

PR #273, opened 2019-10-09
Add an ITensor constructor accepting the desired flux

mtfishman:
@emstoudenmire, does this seem like a reasonable thing to add? It mirrors the `randomITensor` constructor.

@PBluntz needed to make an ITensor filled with zeros with a particular divergence and I realized that it is harder than it should be to do that. The most straightforward way we could think of was `0.*randomITensor(QN(2),i,j,k)` (which is not very nice). Otherwise you have to make a default ITensor with the indices and then set an element to zero, but you would need to know a location in the tensor that has the flux you want which is a bit annoying. Perhaps there is an easier way we are not seeing, but this constructor seems like the most straightforward way.

### DIFF
--- a/itensor/itensor.cc
+++ b/itensor/itensor.cc
@@ -71,6 +71,13 @@ ITensor(Cplx val)
         }
     }
 
+ITensor::
+ITensor(QN q, IndexSet const& is)
+  :
+  is_(std::move(is))
+  {
+  store_ = newITData<QDenseReal>(is,q); 
+  }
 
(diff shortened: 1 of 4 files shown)


emstoudenmire:
Yes I think this is very reasonable to add, and I like how it puts the QN argument first.

(Btw not sure what's going on with Jenkins. We should take a look at its setup again soon to make sure it's useful for us, and possibly consider switching to a different system like Travis that could be nicer.)

PR #275, opened 2019-10-11
Add DMRG Ising test with QNs

mtfishman:


emstoudenmire:
I'm good with pulling this - any part we should look at or discuss? Thanks -

mtfishman:
I think it is pretty straightforward, I was just waiting for tests to run before merging it. This is testing a new feature I added to `SpinHalf` to optionally conserve parity symmetry instead of (or along with) `Sz` symmetry, which is useful for the transverse field Ising model.

PR #276, opened 2019-10-17
Fixed possible memory leakage

mingruyang:
and removed unnecessary generation of temporary itensor objects.

### DIFF
--- a/itensor/iterativesolvers.h
+++ b/itensor/iterativesolvers.h
@@ -1035,7 +1035,7 @@ applyExp(BigMatrixT const& A,
             // Arnoldi: Modified Gram-Schmidt
             for(size_t j = 1; j <= actual_maxiter; ++j)
                 {
-                ITensor p;
+                auto& p = V[j];
                 A.product(V[j-1],p);
                 for(size_t i = 1 ; i <= j; ++i)
                     {
@@ -1051,7 +1051,7 @@ applyExp(BigMatrixT const& A,
                     break;
                     }
                 H(j,j-1) = s;
-                V[j] = (p*=(1.0/s));
+                p *= (1.0/s);
                 }
             }
         else
@@ -1059,7 +1059,7 @@ applyExp(BigMatrixT const& A,
             // Lanczos
             for(size_t j = 1; j <= actual_maxiter; ++j)
                 {
-                ITensor p;
+                auto& p = V[j];
                 A.product(V[j-1],p);
                 if(j != 1) p -= H(j-2,j-1)*V[j-2];
                 H(j-1,j-1) = (dag(V[j-1])*p).eltC();
@@ -1074,7 +1074,7 @@ applyExp(BigMatrixT const& A,
                     }
                 H(j-1,j) = s;
                 H(j,j-1) = s;
-                V[j] = (p*=(1.0/s));
+                p *= (1.0/s);
                 }
             }
 
--- a/itensor/iterativesolvers.h
+++ b/itensor/iterativesolvers.h
@@ -962,9 +962,9 @@ applyExp(BigMatrixT const& A,
     auto maxiter_ = args.getSizeT("MaxIter",30);
     auto errgoal_ = args.getReal("ErrGoal",1E-12);
     auto debug_level_ = args.getInt("DebugLevel",-1);
-    int ideg_ = args.getInt("PadeApproxDeg",6);
     auto orthot_ = args.getBool("IsHermitian",false);//true: using Lanczos; false: using Arnoldi
-
+    auto ideg_ = args.getInt("PadeApproxDeg",6);
+    
     size_t maxsize = A.size();
     auto actual_maxiter = std::min(maxiter_,maxsize);
     if(debug_level_ >= 2)
@@ -986,7 +986,8 @@ applyExp(BigMatrixT const& A,
     //Real Anorm = A.norm();// infinite norm of A, not efficient to calculate.
     ITensor temp;
     A.product(phi,temp);
-    Real Anorm = eltC(dag(phi)*temp).real();
+    Real Anorm = norm(temp);
+	//Real Anorm = eltC(dag(phi)*temp).real();
 	
     int maxrej = 10;// maximum allowable number of rejections at each step
     auto break_tol = 1E-14;// break tolerance when doing orthogonalization
@@ -1013,7 +1014,7 @@ applyExp(BigMatrixT const& A,
     t_new = std::ceil(t_new/s)*s;
 
     auto& w = phi;
-    // hum = max||exp(-sA)||, where s in [0,t].
+    // hump = max||exp(sA)||, where s in [0,t] when t > 0 or [t,0] when t < 0.
     // It measures the conditioning of the matrix exponential. 
     // Well conditioned if hump = 1.
     Real hump = vnorm;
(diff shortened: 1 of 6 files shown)


mtfishman:
Hi Mingru,

Does this help to improve the performance?

-Matt

mingruyang:
Hi Matt,

This does not decrease the calculation time a lot. I just wanted to decrease the memory usage. Since my TDVP stops somehow when the bond dimension only grows to about 200 and the memory usage is not normal. But I haven't checked if this push fix the memory problem.

I did several tests and found that the criterion to determine how many time-step adjusting to be used might be different in my implementation from Jutho's. So sometimes the number of times to do matrix-vector multiplication is different, and this results in the different time cost in the comparison. We might need to also check the .product function of LocalOp in ITensor c++ to make sure it is efficient. Since the orthogonalization dominates the time cost.

In my push, I did several changes which I think is more reasonable to do.

1. I removed the expMatrixApply but just calculate the exp of the matrix instead. Since I realized   we don't need to actually multiply the matrix with a unit vector but just need to select the first column of the matrix.

2. I added a function to do expHermitian for matrix and use it in applyExp when IsHermitian = true.

3. I removed the energy calculation in the applyExp, since this is not the responsibility of this function. We can turn it on or off in the TDVP routine by using an observer. The return value of applyExp is now the condition number hump. We can make a struct of "info" to be the return value in the future.

4. I changed the way to calculate the norm of the BigMatrixT to the way Jutho used.

Let me make several comments on the second point.

I found a case where the krylov subspace method and expPade can fail. My test is on a 1000*1000 uniformly distributed random matrix and made it to Hermitian. I calculated exp(tA) where t = 0.1. Both Jutho and my implementation of krylov will give explosively different answer from the expm method in Julia and expHermitian method in ITensor. expPade also fails in this case. But I guess this case is rare and unphysical, so we need not to worry about that. (For random matrix of normal distribution, they are all fine.)

I have another suggestion in the documentation of ITensor. I found that the type of distribution 
 of the randomITensor is not mentioned. But I guess it is uniform distribution? I think it would be better to tell the user in the documentation. And I think it will also be useful to generate random matrix of normal distribution.

mingruyang:
In my tests, I use t = 0.1, and want to exponentiate exp(tA), where A is a random Hermitian matrix.

mingruyang:
Just made another commit to accept numCenter args in sweepnext so that we can compress the TVDP code further.

mtfishman:
Yes, ITensors are filled with a uniform distribution. They use the `quickran()` function which is implemented [here](https://github.com/ITensor/ITensor/blob/1011a6f9c56eccf60b117920fc06d6f296ca8e58/itensor/detail/algs.h#L96). That would be good to mention in the docs, and possibly give an option for a normal distribution (though so far for cases we use it for, like initializing an MPS product state or testing out an algorithm with random inputs, the specific distribution shouldn't matter so much).

Thanks for looking into this, the changes in this pull request sound reasonable (whether or not they fix the performance/memory problems). Interesting to hear that the `expPade` is having such a large influence on the convergence of the algorithm. @alexwie, what algorithm do you use for dense exponentiation?

mtfishman:
Did you mean to close this?

mingruyang:
I just realized I made a mistake in testing the expPade so I modified my previous statement. For random matrix of normal distribution, expPade is fine and accurate. The only problem happens in random matrix of uniform distribution, which also appears in the krylov method (both in my implementation and Jutho's).

itensor/mps/sweeps.h
```diff
@@ -486,24 +486,12 @@ operator<<(std::ostream& s, const Sweeps& swps)
     }
 
 void inline
-sweepnext(int &b, int &ha, int N, int min_b = 1)
+sweepnext(int &b, int &ha, int N, Args const& args = Args::global(), int min_b = 1)
```
mtfishman: Should the `Args` come last? This may be breaking if someone input `min_b`.

We may want to make `min_b` input as an arg, and then have a wrapper function for backwards compatibility. For example:
```C++
void inline
sweepnext(int &b, int &ha, int N, Args const& args = Args::global())
  { // The new version
  }

void inline
sweepnext(int &b, int &ha, int N, int min_b)
  { // Creates an Arg, adds min_b to it, and calls the other sweepnext
  }
```

mingruyang:
I just checked the condition number of the random matrix which is uniform distributed. It has a eigenvalue which is much larger than the others. So I think adding a precondition process will improve the accuracy both in my and Jutho's implementation of Krylov subspace method to do matrix exponential and my implementation of expPade.

itensor/mps/sweeps.h
mingruyang: But what if someone input args but not min_b?

Yes, I think have a wrapper function is a good idea.

itensor/mps/sweeps.h
```diff
@@ -498,6 +499,20 @@ sweepnext(int &b, int &ha, int N, Args const& args = Args::global(), int min_b =
         }
     }
 
+//void inline
+//sweepnext(int &b, int &ha, int N, int min_b = 1)
```
mtfishman: For backwards compatibility, we should keep this version of `sweepnext`, but change it to `int min_b` instead of `int min_b = 1`. The point is to have backwards compatibility for someone who had called `sweepnext(1,1,20,2)` (which in this pull request right now would result in an error). If you call `sweepnext(1,1,20)` it would call the new version.

itensor/iterativesolvers.h
```diff
@@ -1027,20 +1028,20 @@ applyExp(BigMatrixT const& A,
         Real t_step = std::min(t_out-t_now,t_new);
 
         auto V = std::vector<ITensor>(actual_maxiter+1);// storage for the bases of the Krylov subspace
-	auto H = CMatrix(actual_maxiter+2,actual_maxiter+2);// storage for the projected matrix
+        auto H = CMatrix(actual_maxiter+2,actual_maxiter+2);// storage for the projected matrix
 
-        V[0] = 1.0/beta * w;
+        V[0] = (w *= 1.0/beta);
```
mtfishman, line 1033: What was the reason for this change? I think the first version is fine (and more readable).

itensor/iterativesolvers.h
```diff
@@ -1163,10 +1174,10 @@ applyExp(BigMatrixT const& A,
 
         // Update w = beta * V * exp(sgnt * t_step * H) * e1
         mx = mbrkdwn + std::max(0,k1-1);
-        w = V[0]*F(0);
+        w = (V[0]*=F(0,0));
         for(int i = 1; i < mx; ++i)
             {
-            w += V[i]*F(i);
+            w += (V[i]*=F(i,0));
```
mtfishman, line 1180: Can this just be `w += V[i]*F(i,0)`? Also a similar question for the line right before the loop.

mingruyang: I thought *= will avoid creating a temporary ITensor?

mtfishman: Hmm yes I suppose here it might. Though maybe it is better written as
```C++
V[i] *= F(i,0);
w += V[i];
```
just for clarity.

But I think the change before the loop is not necessary.

mtfishman:
I am going to merge this since the changes seem reasonable and I would like to start working on this code.

PR #278, opened 2019-10-31
Add some more Matrix operations, extend expMatrix to Real inputs

mtfishman:
When trying to make @alexwie's Krylov method for exponentiating an operator work with the latest version of ITensor, I noticed the `expGeneral` didn't work with Real inputs. This makes it work with Real/mixed inputs, as well as changes the interface a little bit.


emstoudenmire:
Glad to see the matrix code getting these upgrades.

PR #280, opened 2019-11-04
Always check for proper divergence when adding ITensors (not just in …

mtfishman:
…debug mode)


mtfishman:
Suggested by Daniel. This seems like something that should always be checked (and shouldn't add much cost). @emstoudenmire let me know if you think this is reasonable.

emstoudenmire:
Yes, I think it's totally reasonable to add checks for correctness as long as we can see they are nearly zero-cost. Probably the reason we didn't have this is I was once concerned about the cost, but I doubt it's ever a big cost except when adding small tensors which is fast anyway.

PR #284, opened 2019-11-06
Use std::remove instead of system("rm ...")

kyungminlee:


emstoudenmire:
Thanks for the PR. Does the code do the same thing before and after this change? My understanding was that std::remove is for removing elements from a data structure, such as an array, which is in RAM, and not for issuing system commands such as "rm" ? But maybe this is a different use case?

emstoudenmire:
Here are the docs on std::remove I found:
https://en.cppreference.com/w/cpp/algorithm/remove

mtfishman:
There is also:

https://en.cppreference.com/w/cpp/io/c/remove

That is very confusing, maybe it is an old c function?

kyungminlee:
Yes it's a function in stdio.h, and is overloaded with std::remove in algorithm. This PR should be useful when using MSVC to build itensor (with other minor changes here and there).

There is also the std::filesystem::remove and remove_all in c++17, but I'm not sure if using filesystem is a good idea since it requires linking a separate library in many compilers.

emstoudenmire:
Ok thanks for clarifying. Looks good to go!

PR #285, opened 2019-11-08
QR Decomposition Support

Jack-Kemp:
Add basic QR decomposition support to ITensor using the LAPACK wrappers. The interface and implementation are effectively copied from the SVD interface (without truncation).

My personal use case is DMT, but looking at the message boards QR decomposition has come up several times before.


emstoudenmire:
Thanks for a very high quality pull request! We certainly needed to have this and thanks for doing the hard work of building up all of the layers.

Matt and I will review it and let you know (we will definitely pull, just may request a few small changes first).

Jack-Kemp:
Great, glad it may be of use. The code definitely needs some eyes on it. I haven't used it in a real world problem yet.

When I have time i will try and see why the unit test build is failing in Jenkins -- obviously I ran the the tests locally it but clearly I missed something!

Jack-Kemp:
The problem appears to be that zungqr needed to be called as LAPACKE_zungqr for the reference LAPACK, which I have and could test now works. 

I see from the code that ACML may also require special calls -- unfortunately I don't have easy access to a system with ACML (is it not end-of-life?), and I see that Jenkins does not seem to test this either. If someone who knows more about ACML could advise that would be great.

itensor/decomp.cc
```diff
@@ -925,4 +925,185 @@ expHermitian(ITensor const& T, Cplx t)
     return prime(U)*d*dag(U);
     }
 
+void
+qr_decomp(ITensor const& AA,
```
mtfishman: Could we just name it `qr`?

itensor/decomp.cc
```diff
@@ -925,4 +925,185 @@ expHermitian(ITensor const& T, Cplx t)
...
+	
+	auto qL = Index(move(Liq),qI.dir(),internaltagset);
+	auto rL = qL;
+	rL.setDir(rI.dir());
+	rL.setTags(internaltagset);
+	auto Qis = IndexSet(qI,dag(qL));
+        auto Ris = IndexSet(rL, rI);
+
+        auto Qstore = QDense<T>(Qis,QN());
+        auto Rstore = QDense<T>(Ris,QN());
```
emstoudenmire: There will be some cases where the flux (total QN) of the original tensor is not equal to zero, in which case R should absorb this flux. (Q should always be flux zero since it is unitary i.e. just a change of basis.) 

It would be great if you could add a unit test which tests a non-zero flux tensor T and checks that its QR decomp works (I think it won't unless you propagate T's flux here to R).

Jack-Kemp:
A further complication with QR with ITensors with QNs is that R isn't currently strictly upper triangular depending on the non zero blocks of T the input tensor. E.g. if 

```
T = 
ITensor ord=2: 
(5|id=752) <Out>
  1: 3 QN({2})
  2: 2 QN({-1})
(6|id=413) <Out>
  1: 2 QN({2})
  2: 4 QN({-1})
QDense Real {2 blocks; data size 16}
Block: 2 1
...
Block: 1 2
...

R = 
ITensor ord=2: 
(5|id=389|QR,Link) <Out>
  1: 3 QN({2})
  2: 2 QN({-1})
(6|id=413) <Out>
  1: 2 QN({2})
  2: 4 QN({-1})
QDense Real {2 blocks; data size 16}
Block: 2 1
(4,1) -0.908769
(4,2) -0.353330
(5,2) 0.0746456
Block: 1 2
(1,3) -1.283717
(1,4) -0.851914
(2,4) 0.3548210
(1,5) -1.347467
(2,5) 0.3534045
...
```

Here block 2 1 of R is actually lower triangular, as it can't be upper triangular by nature of the position of the block, and so the entire tensor R isn't upper triangular. I'm not sure if this is fixable given we must have T = Q*R for Q unitary, but I may well be missing something obvious.

emstoudenmire:
Good point. I think it will be just fine to allow this behavior, and it may be unavoidable when combined with the constraints on the fluxes of the ITensors + the way the indices are laid out ahead of time.

I see QR as having two main advantages, with the first one being much more important in tensor networks than the second:
1. it is fast to compute and computes a factorization with one matrix / tensor being unitary
2. R is triangular so can be used to back-solve linear equations

So I think that giving up #2 is ok for QN sparse tensors, since we mainly want QR to take advantage of #1. (Unless you have a specific use case in mind which requires #2.)

Jack-Kemp:
In fact I was mistaken and it is possible to get R upper triangular by appropriate permutations and splitting in two of the blocks in the internal index shared by Q and R without changing the unitarity of Q, or the external indices. This has the advantage that one can then take the not complete, 'thin' QR decomposition (e.g. [Wikipedia](https://en.wikipedia.org/wiki/QR_decomposition)).

I have added two arguments: Complete, which controls whether the QR is thin or not, and UpperTriangular, which controls whether to bother to do the permutations. (Complete and not UpperTriangular) should give the previous behaviour. The default currently is both true -- which matches Matlab, but not numpy or Mathematica which both give thin QRs as default. Apologies for the large change after submitting for review but I hadn't considered the complications of the block structure.

As for a use case of upper triangular R, DMT technically requires it -- e.g. equation 18 from https://arxiv.org/pdf/1707.01506.pdf , though admittedly I haven't considered whether that analysis immediately generalises when conserving QNs.

Regardless of any of that some of the unit tests should really check the sanity of the Q and R tensors -- I'd appreciate any hints as to the cleanest way to check an ITensor is upper triangular.

emstoudenmire:
Hi, sorry for moving slowly on this. I'm looking forward to having it in the library.

That's really great you figured out how to get R to be upper triangular even for the QN block-sparse case. I can definitely imagine it's possible, but it wasn't totally obvious.

So I'd only suggest one change at a high-level, which is let's make the QR "thin" by default. The reason is that in a context like tensor networks, which as you know are all about *low* rank structures, one wants to compress indices as much as possible, especially when it can be done in a lossless way. One of the main uses of QR, as you know, is bringing MPS into canonical form and without the thin QR the bond dimension would grow I believe.

Finally, regarding checking the upper-triangularity in a unit test, a simple although slightly inefficient way would be to just get each element of R that's expected to be zero and check that it is so. As you may know, it's ok to call T.elt(...) to retrieve elements of a block sparse tensor that are "missing" blocks; the code just returns zero for those. As long as the test case you do isn't too large, the slight inefficiency of the .elt method won't be a big deal.

Best regards,
Miles

Jack-Kemp:
Agreed about the default being thin QR -- I have changed this. I also fixed a bug where using the complete option was not giving fully complete due to vanishing blocks in A, updated the unit tests as discussed above and accounted for the changes in the getBlock function which seems to have occurred since I last merged.

emstoudenmire:
Great - this looks ready to go. Thanks again for the high quality code!

One small improvement I will probably make after merging it is to make the lower-level QR function always return an R which has non-negative diagonal elements. Doing so is sufficient to make the Q unique (maybe also one has to assume the matrix being decomposed does not have linearly dependent columns too?). Someone we were talking to about an MPO algorithm commented that they found this was important, so it might be important for DMT too.

PR #288, opened 2019-11-13
WIP: new QDense design

mtfishman:
Here is an initial prototype for a new QDense storage design and contraction algorithm. For a 4x4 Hubbard model at U=2 with particle number, spin and momentum around the cylinder conserved, the timings are the following:

This pull request
```
    vN Entropy at center bond b=8 = 3.019814122626
    Eigs at center bond b=8: 0.1677 0.1677 0.1597 0.1597 0.0138 0.0138 0.0138 0.0138 0.0096 0.0096 
    Largest link dim during sweep 5/5 was 7341
    Largest truncation error: 4.5789e-09
    Energy after sweep 5/5 is -18.876818002901
    Sweep 5/5 CPU time = 18m, 25s (Wall time = 2m, 26.5s)
```

Current v3
```
    vN Entropy at center bond b=8 = 3.019814122626
    Eigs at center bond b=8: 0.1677 0.1677 0.1597 0.1597 0.0138 0.0138 0.0138 0.0138 0.0096 0.0096 
    Largest link dim during sweep 5/5 was 7341
    Largest truncation error: 4.5789e-09
    Energy after sweep 5/5 is -18.876818002901
    Sweep 5/5 CPU time = 29m, 42s (Wall time = 7m, 52.9s)
```

The major speedups are from an improved algorithm for finding the output blocks from the input blocks (instead of from the flux), caching the information about which blocks contract to which blocks when performing the contraction, as well as storing the cartesian indices of the block locations instead of the linear indices to avoid unnecessary integer divisions.

Here are some things that need to be done before merging:

- [x] Complete the write-to-disk and read-from-disk functionality for the new QDense storage
- [ ] Decide how to deal with backwards compatibility of saved files.
- [x] Write the code for setting elements of a "block deficient" QDense storage, i.e. setting the element of a block that could exist according to the flux but doesn't because it resulted from a QDense*QDense contraction that didn't create that block.
- [x] Make sure that functions like `.fill` for QDense storage fills any blocks that should exist according to the QN flux of the tensor (unfortunately, this would slow down all calls to `.fill` to some extent, since it will need to check if all of the blocks exist).
- [x] Optimize QDense+QDense (for example, right now the logic of merging the blocks doesn't use the fact that they are already sorted, and sorting two sorted blocks can be optimized).

Thanks a lot to Alex Wietek for pointing out the performance issue when many QNs are used and helping with profiling to pin down slow parts of the code!


itensor/itdata/qdense.cc
```diff
@@ -444,7 +483,78 @@ doTask(PlusEQ const& P,
...
+
+    // Store the blocks of the output
+    // TODO: optimize this to merge and remove repeats
+    //       at the same time, assuming A.offsets and 
+    //       B.offsets are already sorted
+    auto Cblocks = Blocks();
+
+    for(auto& aio : A.offsets)
+        {
+        Ablock = aio.block;
```
mtfishman: Assignments like this turned out to be surprisingly expensive when performed many times, this is a reminder to just use `aio.block` directly instead of assigning to `Ablock`.

emstoudenmire: These are great finds. What are the main methods you’re using to profile? (Just so I can know what’s working the best.)

mtfishman: We are using gperftools, Nils has a good set of instructions on how to use it from the C++ bootcamp:

https://github.com/Wentzell/Cpp_and_Tools_Bootcamp/blob/master/Tools/Profiling/rng_example/commands

It really helped us identify the bottlenecks. One small thing is that it is important to run with a single MKL thread (`export OMP_NUM_THREADS = 1`), otherwise you get a lot of extraneous information about the threads in the profile data.

emstoudenmire: Interesting. Glad I asked because I had tried gperftools a while back and was not that impressed, and one reason is that I kepy seeing all of these things about threading in the data. So I needed to know that trick at the time!

emstoudenmire: I'm surprised by the way about the `Ablock = aio.block` line above being a problem. Aren't both of those variables just integers? I would think that (a) that would be very cheap to just copy an integer and (b) the compiler could likely optimize that assignment away. 

But if changing that line does give a measureable speedup then you can't argue with the results! Optimizations can be non-intuitive.

mtfishman: In the current design, `aio.block` is an `IntArray` of the cartesian coordinates of the block. `Ablock` is also an `IntArray` (in the code, I made `Block` a typedef for `IntArray`). So the cost was assigning `IntArray`s to each other.

emstoudenmire: Great - makes more sense about the cost of that!

mtfishman: Still surprising how much of a cost they were, since they are very small. We definitely wouldn't have caught it without the profiler.

mtfishman:
I just added functionality for creating Dense and QDense storage with undefined data, which speeds up the constructors a lot. The notation is:
```C++
Dense<Real>(undef,10);
```
to make a length 10 Dense storage with undefined data. The original constructor `Dense<Real>(10)` still makes storage filled with zeros.

In this branch, the new constructor is used in the Dense contraction code, which I have seen speeds up Dense tensor contraction. The plan is to use the new constructor in the QDense contraction code as well.

mtfishman:
The latest timings on this branch, after changing the QDense storage to allocate uninitialized memory in the contraction code (instead of zeros):
```
    vN Entropy at center bond b=8 = 3.019814122626
    Eigs at center bond b=8: 0.1677 0.1677 0.1597 0.1597 0.0138 0.0138 0.0138 0.0138 0.0096 0.0096 
    Largest link dim during sweep 5/5 was 7341
    Largest truncation error: 4.5789e-09
    Energy after sweep 5/5 is -18.876818002901
    Sweep 5/5 CPU time = 16m, 46s (Wall time = 2m, 0.9s)
```
so around a 20% speedup from the original pull request.

emstoudenmire:
This is all really great and overdue. I should have done some of this profiling before, but partly it was because I ran into the grpof threading issue (and also I did catch the ldiv one using timers but sort of didn't believe those results or have time to experiment with switching the block representation). 

Once you get the setting of elements and fill handled, I'd say this is ready to go. Will be great to have the same logic going into the Julia version.

mtfishman:
@emstoudenmire I believe this is ready to go in terms of functionality that needs to be added.

I think the only thing left to think about is the proper way to merge this, since technically it is a breaking change to reading and writing QDense storage.

We haven't been doing a good job tagging minor releases, but I was think that we could tag the current commit of `v3` as `v3.0`, and then when we merge this branch it will be considered `v3.1`. This would help mark when the breaking change will happen, and we can make some functionality to read pre-v3.1 ITensor files.

emstoudenmire (approved):

emstoudenmire:
Hi Matt, this is great. Glad to see the really big, concrete speedups without much increase, if any, in code complexity.

For the versioning, I think we should be able to go back and tag a v3.0.0 from when we first made the v3 branch (should have done that already! thought we had). Then I think we should tag this one as v3.1.0.

PR #291, opened 2019-11-18
Make build process a bit more robust

mtfishman:
This issue was brought up by Nils when he was trying to build ITensor in a Dropbox directory that contained spaces and parentheses. The current build process errors if the path that ITensor is living in has spaces or parentheses (or both) because they weren't being escaped. Surrounding the THIS_DIR variable with single quotes (but not double quotes) seems to fix this problem.


mtfishman:
Looks like this caused a problem for building the tests, I need to look into it more.

mtfishman:
The latest commit fixes the build system if the path contains parentheses. However, it seems like make has some serious issues when dealing with spaces in file names, so it may be impractical to support that.

emstoudenmire:
Thanks for taking care of this. Seems like a definite improvement, even if it doesn't yet fix the issue with spaces.

PR #294, opened 2019-11-21
Faster applyExp

mtfishman:
Here is a simplified, faster version of applyExp. For now, support is removed for non-Hermitian matrices.

### DIFF
--- a/itensor/iterativesolvers.h
+++ b/itensor/iterativesolvers.h
@@ -69,11 +69,11 @@ gmres(BigMatrixT const& A,
 // It does not compute the matrix exponential explicitly
 // but instead compute the action of the exponential matrix on the vector.
 // After finished, phi -> phi'.
-template <class BigMatrixT>
-Real
+template <typename BigMatrixT, typename ElT>
+void
 applyExp(BigMatrixT const& A,
          ITensor& phi,
-         Cplx t,
+         ElT t,
          Args const& args = Args::global());
 
 //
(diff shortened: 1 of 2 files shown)


mtfishman:
Though for this case, the matrices involved are definitely not uniform distributed random Hermitian matrices since they are tridiagonal (and built with Lanczos, so very specific). In general they will also be quite small (less than 100x100). Perhaps we can keep it in mind and see if we see any problems.

titaschanda:
Question: In the current (as well as the previous) implementation, Lanczos iteration is being done without any reorthogonalization. Usually (i.e., when finding eigenstates), Lanczos without reorthogonalization gives inaccurate results (spurious Ritz values etc.). Is this reorthogonalization step unnecessary for Lanczos exponentiation and should not incur much error?

In my own implementation (of TDVP, which uses similar exponentiation like the current one), I am seeing changes between the situations of "full reorthogonalization" and "no reorthogonalization"  at third to fifth decimal places (in entropy for example) depending on the time scale. On the other hand, situations without reorthogonalization are 20 to 30% faster as expected, and hence preferable if incurred errors are not much.

titaschanda:
Update after some tests: The results without reorthogonalization become unreliable for following scenarios (of TDVP):

1. Energy eigenvalues of the Hamiltonian have large absolute values (eg. Bose-Hubbard model etc.) i.e., H_{effective} has large matrix-norm, and/or
2. The time-step \tau is large (say, ~ 0.1), and/or
3. Total elapsed time is large (say, ~100 or larger).

In such cases, "with full re orthogonalization" and "without reorthogonalization" produces visibly different results. I just saw that this issue was also mentioned in arXiv:1901.05824. I will try to make a pull request adding an optional reorthogonalization step.

PR #302, opened 2019-11-23
Fix deprecated UseOrigM error message

Jack-Kemp:
Fix the somewhat confusing typo in the deprecation message.


mtfishman:
Thanks!

PR #309, opened 2019-12-09
"EntropyErrgoal" and Arbitrary Spin

jzhan039:
1. "EntropyErrgoal" is added to DMRGObserver.h.
2. Add sites files for arbitrary integer spin (spinint.h) and half integer spin (spinhalfint.h).

### DIFF
--- a/itensor/all_mps.h
+++ b/itensor/all_mps.h
@@ -43,6 +43,9 @@
 
 #include "itensor/mps/sites/spinhalf.h"
 #include "itensor/mps/sites/spinone.h"
+#include "itensor/mps/sites/spintwo.h"
+#include "itensor/mps/sites/spinint.h"
+#include "itensor/mps/sites/spinhalfint.h"
 #include "itensor/mps/sites/boson.h"
 #include "itensor/mps/sites/electron.h"
 #include "itensor/mps/sites/fermion.h"
--- a/itensor/mps/DMRGObserver.h
+++ b/itensor/mps/DMRGObserver.h
@@ -58,11 +58,17 @@ class DMRGObserver : public Observer
 
     MPS const& psi_;
     Real energy_errgoal; //Stop DMRG once energy has converged to this precision
+    Real S; //von Neumann entropy
+    //Real Renyi2; //2nd order Renyi entropy
+    //Real DEnergy; //energy difference between last two sweeps
+    //vector<Real> EnergyLis; //list of all energies of dmrg sweeps
+    Real entropy_errgoal; //Stop DMRG once entropy has converged to this precision
     bool printeigs;      //Print slowest decaying eigenvalues after every sweep
     int max_eigs;
     Real max_te;
     bool done_;
     Real last_energy_;
+    Real last_entropy_;
     Spectrum last_spec_;
 
     /////////////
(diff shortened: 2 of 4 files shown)


emstoudenmire:
Thanks very much for this! Regarding the entropy errgoal code, the code is good but I think the extra part where it checks if both energy and entropy error goals are set is not needed. This is because the behavior is the same as when just one of them is set, which will exit as soon as one of them is reached. So think the behavior is already covered by looking at them separately only. But please let me know if I am missing a case that you noticed.

Regarding the separate spin int and spin half cases, I think a nicer way to go would be to merge these together as one (if that's easy to do) then when the user passes the spin value, multiply it by two and then round the result to the nearest integer (there should be a std library function for this) to get the integer specifying the spin precisely.

Alternatively the user can pass an integer where 1 means spin 1/2, 2 means spin 1, 3 means spin 3/2 etc. Both options could be supported by giving them different names.

jzhan039:
> Thanks very much for this! Regarding the entropy errgoal code, the code is good but I think the extra part where it checks if both energy and entropy error goals are set is not needed. This is because the behavior is the same as when just one of them is set, which will exit as soon as one of them is reached. So think the behavior is already covered by looking at them separately only. But please let me know if I am missing a case that you noticed.
> 
> Regarding the separate spin int and spin half cases, I think a nicer way to go would be to merge these together as one (if that's easy to do) then when the user passes the spin value, multiply it by two and then round the result to the nearest integer (there should be a std library function for this) to get the integer specifying the spin precisely.
> 
> Alternatively the user can pass an integer where 1 means spin 1/2, 2 means spin 1, 3 means spin 3/2 etc. Both options could be supported by giving them different names.

Sorry, I was planning to stop sweeps when both of them are met. Maybe sometimes we need both energy and entropy to be accurate.

For the spin int and spin half, actually the "spinhalfint.h" works for both integer spin and half integer spin. One just need to set "DSmax" to be twice the spin, just as you said in the end. Thanks.

emstoudenmire:
Yes, that makes more sense about leaving that code to handle the "and" case, so I agree with that change.

I'll go ahead and merge, but we may make some edits to the new site sets and just keep the spinhalfint one if you don't mind.

jzhan039:
> Yes, that makes more sense about leaving that code to handle the "and" case, so I agree with that change.
> 
> I'll go ahead and merge, but we may make some edits to the new site sets and just keep the spinhalfint one if you don't mind.

OK, thanks. I agree to just keep "spinhalfint".

emstoudenmire:
Great - ok I just modified the code slightly, renaming some of the arguments and object names to be more general.

emstoudenmire:
Documented here: http://itensor.org/docs.cgi?vers=cppv3&page=classes/customspin

jzhan039:
> Documented here: http://itensor.org/docs.cgi?vers=cppv3&page=classes/customspin

Great, thanks. Maybe the sentence "...representing a spin with a value S/2 where the user can specify S (or 2S if one prefers)." is a little confusing. You are saying the spin value is S, S can be integer or half integer, users can either specify 2S or S, right? In the code, specifying 2S = 3, or S=1.5 are both spin-3/2.

emstoudenmire:
Thanks - my mistake! There were actually a few mistakes on that page. Could you please take a look now and see if it looks ok to you?

jzhan039:
> Thanks - my mistake! There were actually a few mistakes on that page. Could you please take a look now and see if it looks ok to you?

Yes, it looks OK to me. Thanks.

PR #313, opened 2019-12-21
Position bugfix

emstoudenmire:
Fix for bug found by Alexios M. The bug was caused by the logic of the position method setting LeftTags going into orthMPS (and thus into svd) to equal the default right tags.

By the way, there is a check within svd for this situation which fails, because TagSets can easily end up with negative prime level versus zero prime level depending on how they are constructed (from Args versus the TagSet constructor). Even though they are interpreted as both having prime level zero, the operator== comparison treats them as different.

### DIFF
--- a/itensor/mps/mps.cc
+++ b/itensor/mps/mps.cc
@@ -663,7 +663,14 @@ position(int i, Args args)
             // Store the original tags for link b so that it can
             // be put back onto the newly introduced link index
             auto original_link_tags = tags(linkIndex(*this,b));
-            orthMPS(ref(b),ref(b+1),Fromright,{args,"LeftTags=",original_link_tags});
+            args.add("LeftTags=",original_link_tags);
+            if(original_link_tags == TagSet("Link,V,0"))
+                {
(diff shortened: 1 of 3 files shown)


itensor/mps/mps.cc
```diff
@@ -663,7 +663,14 @@ position(int i, Args args)
             // Store the original tags for link b so that it can
             // be put back onto the newly introduced link index
             auto original_link_tags = tags(linkIndex(*this,b));
-            orthMPS(ref(b),ref(b+1),Fromright,{args,"LeftTags=",original_link_tags});
```
emstoudenmire, line 666: The actual bugfix starts here (the rest are small cosmetic edits)

titaschanda:
In this regard, I want to add that if all the link tags on an MPS is same (or some of them atleast) then position() automatically changes the id's of such link indices for QN non-conserving dense tensors. As a results, psi(b) and psi(b+1) are left with indices with different id's that should connect them. For QN conserving tensors, it is not a problem.

PR #315, opened 2020-01-06
Fix the wrong definition of avnorm in applyExp

mingruyang:
According to [page 17 of this paper](https://epubs.siam.org/doi/abs/10.1137/0729014), `avnorm` which is used for error estimate should be Av_{m+1} instead of Av_m.

### DIFF
--- a/itensor/iterativesolvers.h
+++ b/itensor/iterativesolvers.h
@@ -967,114 +967,364 @@ assembleLanczosVectors(std::vector<ITensor> const& lanczos_vectors,
 
 template<typename BigMatrixT, typename ElT>
 void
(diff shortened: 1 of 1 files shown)


mtfishman:
Hi Mingru,

Sorry for the slow response. Have you seen that this change makes a difference in calculations you are doing (not doubting that it is the correct thing to do, just curious)?

Cheers,
Matt

mingruyang:
Hi Matt,

I only tested it with the sample code in the TDVP repo. Actually I didn’t see any difference in the results but it runs slightly slower.
I think it doesn’t matter in the current setting but might affect the time step adjusting.

Best,
Mingru

PR #321, opened 2020-01-20
Add (full) reorthogonalization to applyExp

titaschanda:
Lanczos iteration usually results in loss of orthogonalization in Krylov vectors, that may cause accuracy issues in applyExp. Reorthogonalization process solves that problem.


titaschanda:
Some minor bugs are there. Will update later.

PR #324, opened 2020-01-23
Quickran fix potential overflow

mtfishman:
Nils pointed out from a sanitizer check that quickran might have an integer overflow issue. I've extended the integer types to long to (hopefully) fix this, though it is hard to check since it is based on a generated process id.

### DIFF
--- a/itensor/detail/algs.h
+++ b/itensor/detail/algs.h
@@ -98,7 +98,7 @@ quickran()
     auto res = 0.0;
     while(res == 0.0)
         {
-        long& seed = seed_quickran(0);
+        long seed = seed_quickran(0);
         long im = 134456;
         long ia = 8121;
         long ic = 28411;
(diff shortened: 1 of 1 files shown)


mtfishman:
Note that I've timed the code:

```c++
    auto n = 20000;
    auto i = Index(n,"i");
    auto j = Index(n,"j");
    auto A = randomITensor(i,j);
```

(which calls quickran on each element of the tensor). Before this change, it takes 3.40s on my laptop. Oddly enough, after this change it takes 3.31s. Not sure why it would be any faster (perhaps arithmetic operations with long are more optimized than other integer types?), but anyway it looks like this is good to go now.

emstoudenmire:
Just a guess, but it could be because on a 64-bit architecture, long integers are more natural for the CPU to handle. So then it could indeed be faster. Thanks for taking a look at this!

PR #330, opened 2020-02-11
Edit Jenkins to remove -fconcepts from osx build

mtfishman:


mtfishman:
I'm not sure why Jenkins is having trouble running that command, it works for me locally.

@dylex, do you know what the problem might be? We are trying to modify the osx build to remove the `-fconcepts` flag (since g++ is an alias for clang++ on macos, and clang++ is giving us errors when the `-fconcepts` flag is used).

dylex:
The built-in OSX sed does not support `-i` without an extension (and generally has quite different syntax from GNU sed, along with many other commands -- a constant source of annoyance on OSX). This should fix it.

mtfishman:
Great, thanks for your help! I figured it was something like that.

PR #334, opened 2020-02-29
Relpace orthog in SVD with QR

### LINKED ISSUE #319: Replace orthog with QR
In the custom recursive SVD algorithm in ITensor, the orthogonalization step is currently done with `orthog`, which appears to be taking up most of the time of the total SVD (based on profile data). It may be better to use LAPACK's QR implementation instead.

Jack-Kemp:
Fixes #319 .

As I was interested and I implemented the QR I thought I might as well have a first attempt at this. Replaces orthog with QR in the ITensor SVD implementation and adds a "PositiveDiagonal" argument to QR to force the diagonal of R to be positive, as @emstoudenmire found to be necessary.

Currently passes the unit tests and very initial benchmarking (the orange line in graphs below -- I'll run the code for better averaging and dimension when I have time) suggests it runs faster and more accurately than before, but obviously should be looked over closely before merging by someone who understands the algorithm properly!

image

image

### DIFF
--- a/itensor/tensor/algs.cc
+++ b/itensor/tensor/algs.cc
@@ -509,7 +509,7 @@ SVDRefImpl(MatRefc<T> const& M,
 
     //Form 'density matrix' rho
     Mat<T> rho,
-           Mconj;
+      Mconj, tempV, R;
     if(isCplx(M)) 
         {
         Mconj = conj(M);
@@ -522,13 +522,6 @@ SVDRefImpl(MatRefc<T> const& M,
 
     //Diagonalize rho: evals are squares of singular vals
     diagHermitian(rho,U,D);
-
-    for(auto& el : D)
-        {
-        if(el < 0) el = 0.;
-        else       el = std::sqrt(el);
-        }
-
 
     //Put result of Mt*U==(V*D) in V storage
     if(isCplx(M))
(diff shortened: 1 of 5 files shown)


emstoudenmire:
Hi Jack, this is great! Thanks for taking the initiative on this. 

The only part of the new code that might need to be changed is the part where you sort the diagonal of R. I believe this step should be unnecessary, because the matrix going into the QR step is the singular values times V, and the singular values matrix should already be sorted by the fact that U was found from digonalizing a Hermitian positive semi-definite matrix, and the diagonalization routine returns the eigenvalues in sorted order.

It might be necessary to pass the negative of rho into diagHermitian, though, to get the proper sorting (largest to smallest instead of the other way around).

Let me know if I'm missing something of course. Does skipping that step (and possibly flipping the sign of rho as mentioned above) still give correct results?

Jack-Kemp:
If I comment out the sorting code it still just works (without - sign on rho) and indeed I had already noticed that the output seemed presorted.

I added the sorting in in an excess of caution because LAPACK QR seems to make no guarantees about the order of R, but I see that if the matrix is already a unitary times presorted diagonal matrix it seems reasonable to assume the QR will always return a sorted R. If we are happy with that then I can certainly push a cleaned up version without the sorting soon.

emstoudenmire:
Hi, yes I believe that’s the argument, namely that the diagHermitian already sorts the eigenvalues such that the column space of the matrix going into the QR corresponds to the singular values in descending order. Then one can recover them by requiring a positive R.

So yes please push a version with the sorting removed, and check that the tests are all passing ok, then I’ll be glad to merge!

Jack-Kemp:
OK, the following commit should work.

[Note: why not store QR tempV into V directly rather than vice versa and then having to do: V &= std::move(tempV); ? Because the QR calls reduceCols on the Q argument and apparently that is not allowed with constant refs such as V, so it fails to build complaining about cv qualifiers.].

itensor/tensor/algs.cc
```diff
@@ -561,7 +554,15 @@ SVDRefImpl(MatRefc<T> const& M,
     //    orthog(columns(V,nlarge,Mr),2);
     //    }
 
-    orthog(V,2);
+    //orthog(V,2);
+
+
+    QR(V, tempV, R, {"Complete", false, "PositiveDiagonal", true});
+    V &= std::move(tempV);
```
emstoudenmire, line 561: Just wanted to point out here I believe this std::move will have no effect, because the &= operator for ITensor::TenRef and related types doesn't currently take advantage of rvalue references and the like. But it's a good idea to try to just move the data of tempV into V, assuming they have the exact same dimensions and strides.

If so, a straightforward way to do this could be to use the V.data() accessor to get a pointer to V's block of memory, and similar for tempV, then do a move on these.

emstoudenmire: Happy to go ahead and merge, though, then work on this as a separate optimization.

Jack-Kemp: Ah, I see. I think someone who is more familiar with ITensor::TenRef class than I could probably more effectively and efficiently implement this optimisation, so I think merging now probably makes the most sense.

emstoudenmire: Ok agreed!

PR #337, opened 2020-03-19
Fix bug in complex SVD LAPACK calls

Jack-Kemp:
Unfortunately there was a  bug in my PR #333 -- namely, the size for one of the matrices in the LAPACK routine was too great, causing it to possibly corrupt memory.

I guess it wasn't caught in the unit tests because the SVD still works: it just causes memory errors down the line rather than straight away.

I noticed there are additional improvements which could be done with all the LAPACK routines (not just the SVDs) with regards to LWORK but I wanted to get this bugfix out straight away.


emstoudenmire:
Thanks for continuing to keep an eye on this new code

PR #338, opened 2020-03-25
Add "dim=" and quotes around tags to Index printing

emstoudenmire:
Brings printing of Index more in line with the Julia version. This style of printing indices is more beginner-friendly, as it teaches or hints what the different outputs mean when printing an Index.


mtfishman:
Looks good to me.

Issue #318, opened 2020-01-17
excited state DMRG doesn't work w/ Electrons


ryanlevy:
Right after entering davidson, itensor crashes with:
```
From line 879, file itensor.cc

div(T1)=QN({"Nf",50,-1},{"Sz",0}) must equal div(T2)=QN() when adding T1+T2
```
This occurs whenever there's an `Electron` in the MixedSiteSet, and for some spin Hamiltonians (see below).

---

To reproduce, take the `mixedspin` sample code and change `using SHalfSOne` to either :
```cpp
using SHalfSOne = MixedSiteSet<SpinHalfSite,ElectronSite>;
```
or 
```cpp
using SHalfSOne = Electron;
```
and add at the bottom of main
```cpp
auto state2 = InitState(sites);
//something generic to start
    for(int i = 1, n = 1; i <= N; i += 2, n += 1)
        {
        if(n%2 == 1) state2.set(i,"Dn");
        else         state2.set(i,"Up");
        }
    for(int i = 2, n = 1; i <= N; i += 2, n += 1)
        {
        if(n%2 == 1) state2.set(i,"Dn");
        else         state2.set(i,"Up");
        }
    auto psi1 = MPS(state2);
    auto wfs = std::vector<MPS>(1);
    wfs.at(0) = psi;
    energy = dmrg(psi1,H,wfs,sweeps,"Quiet");
    printfln("\nGround State Energy = %.10f",energy);
```

This was compiled with ITensor v3 @ 13042c0d

emstoudenmire:
Thanks! Another user reported this issue but we haven't had time to get to it. So I'll try to make it a priority. The test cases are very helpful.

zuozw:
Is this bug fixed? I also need to calculate the excited state.

emstoudenmire:
Just addressed this in PR #341 . Sorry it's been awhile, but if you are able to check this, could you confirm that it's fixed in the latest v3 branch of ITensor? Thanks -

zuozw:
It doesn't work.
After calculating the ground energy, the code throw the message
"From line 479, file /Users/**/ITensor3/itensor/itdata/dotask.h
doTask called on unallocated store pointer
doTask called on unallocated store pointer
Abort trap: 6"


------------------ Original ------------------

emstoudenmire:
Thanks, so now that I fixed the other bug linked in the PR (#341) above, I think I see what's going on in this code and it has a different origin.

[...]

PR #341, opened 2020-04-06
Fix excited-state DMRG bug

emstoudenmire:
This PR (hopefully) fixes a persistent bug with excited state DMRG. The issue found was that addition of QN ITensors was failing when one of the tensors had no blocks, in which case the QN flux was being reported as QN() and then triggering a check that the fluxes were the same.


emstoudenmire:
Related to #318

PR #357, opened 2020-06-01
Fix issue with colliding SVD indices

emstoudenmire:


mtfishman:
We should fix this by just making `uL` and `vL` have different ids, which is already the case for QNs (since in general for QNs they must have different ids, to get the QN flux correct for U,S and V).

emstoudenmire:
Great - I like that fix better. (This issue was breaking a ancilla-method finite T code I was trying out this morning.) I had thought maybe they had the same id’s right now by design, but if not I agree let’s change this. I’ll update the PR with that design.

PR #362, opened 2020-08-17
fixed unwanted range behavior if start > end

chuffa:
Changed the != operator in RangeHelper:
Instead of "!=" use "<" to account for cases where curr_ > end_ from initialization.

Tested the changes so they fix the bug and still give the same output in the other cases.


mtfishman:
Looks good, thanks!

PR #368, opened 2020-10-14
Fix potential integer overflow in getContractedOffsets

kyungminlee:
Change size of `Csize` from `int` to `size_t`.
With the `int Csize` on LP64, contraction throws `std::length_error` at bond dimension of around a few tens of thousands.
Addresses the following issue:
http://itensor.org/support/2604/issue-with-size-of-the-vector-in-large-bond-dimension


mtfishman:
Looks reasonable, thanks!

PR #370, opened 2020-10-22
Remove unnecessary #include from global.h

kyungminlee:
`unistd.h` is not used within `global.h`

### DIFF
--- a/itensor/global.h
+++ b/itensor/global.h
@@ -21,7 +21,6 @@
 #include <cstring>
 #include <memory>
 #include <random>
-#include <unistd.h>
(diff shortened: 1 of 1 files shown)


emstoudenmire:
Thank you Kyungmin. However, do you know if leaving this include is causing any problems such as warnings or significantly slower compile times? If I recall, we added this not because it was always needed, but that on certain rare occasions it was, depending on the platform.

kyungminlee:
You are right. `getpid` is used in `global.cc` for RNG seeding. My bad. I am closing this PR.

PR #379, opened 2021-01-17
fix subscript out of range in autovector.h

shencebebetterme:
In "autovector.h", the length of the std::vector `dat_` is equal to `1+maxi_-mini_+miniloc_`, therefore in the expression `dat_[1+maxi_-mini_+miniloc_]` the subscript definitely goes out of range. 

Previously this issue causes no error because on Linux and on mac `std::vector` does not perform a bound check and returns whatever in the memory as the out-of-range element. The `end()` thus obtained still points to the correct memory, however, the above behaviour is undefined and should be avoided.

The fixes are easy.

### DIFF
--- a/itensor/util/autovector.h
+++ b/itensor/util/autovector.h
@@ -85,15 +85,15 @@ class autovector // returns T() outside range or unassigned
     T*
     begin() { return &dat_[miniloc_]; }
     T*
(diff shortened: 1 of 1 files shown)


emstoudenmire:
Thank you for this fix - it's a welcome improvement!

Issue #360, opened 2020-07-29
error in calling arnoldi function


shencebebetterme:
'rom [this post](http://itensor.org/support/1615/solve-an-eigenvalue-problem-of-non-hermitian-tensor) I can see how to use the arnoldi function to get the largest eigenvalue of a square matrix. However, if I attempt to calculate the first k eigenvalues (k>1) and feed the arnoldi function with a vector of ITensor rather than a single ITensor, the compiler throws a long template error.

The minimal codes to reproduce the error is 
```
#include "itensor/all.h"
using namespace itensor;

class ITensorMap
  {
  ITensor const& A_;
  mutable long size_;
  public:

  ITensorMap(ITensor const& A)
    : A_(A)
    {
    size_ = 1;
    for(auto& I : A_.inds())
      {
      if(I.primeLevel() == 0)
        size_ *= dim(I);
      }
    }

  void
  product(ITensor const& x, ITensor& b) const
    {
    b = A_*x;
    b.noPrime();
    }

  long
  size() const
    {
    return size_;
    }

};

int
main()
  {
  auto i = Index(2,"i");
  ITensor A =randomITensor(i,prime(i));

  auto AM = ITensorMap(A);

  ITensor x = randomITensor(i);
  ITensor y = randomITensor(i);
  std::vector<ITensor> vec = {x,y};

  auto lambda = arnoldi(AM,vec,{"ErrGoal=",1E-14,"MaxIter=",20,"MaxRestart=",5});

  // Check it is an eigenvector
  PrintData(norm((A*x).noPrime() - lambda*x));

  return 0;
  }
```

In the post MattFishman said he also encountered a problem when calculating more than one eigenvalues/eigenvectors, it seems this issue hasn't been fully solved.

mtfishman:
Hi,

It seems like the error you are seeing is when you are checking the eigenvector, since you are doing `lambda * x` but `lambda` is now a vector of the eigenvalues. This code runs:
```
#include "itensor/all.h"

using namespace itensor;

class ITensorMap
  {
  ITensor const& A_;
  mutable long size_;
  public:

  ITensorMap(ITensor const& A)
    : A_(A)
    {
    size_ = 1;
    for(auto& I : A_.inds())
      {
      if(I.primeLevel() == 0)
        size_ *= dim(I);
      }
    }

  void
  product(ITensor const& x, ITensor& b) const
    {
    b = A_*x;
    b.noPrime();
    }

  long
  size() const
    {
    return size_;
[...]

PR #382, opened 2021-02-02
fix arnoldi eigenvalues

shencebebetterme:
The incorrect eigenvalues calculated by `arnoldi`, as observed in #360 , is due to an extra multiplication of matrix A in the "deflation" process. https://github.com/ITensor/ITensor/blob/fb83fde4a07f2122cf56f9a79f8e8ec01c7cad18/itensor/iterativesolvers.h#L809
We want the Hessenberg matrix `Href` to look like <V_i | A - A0 | V_j>  where the subtracted part A0 is constructed from previously calculated eigenvalues and eigenvectors. The first part is still obtained as <V_i | V_j+1> = <V_i | A | V_j>, however in the codes the seconed part looks like <V_i | A0 | V_j+1> =  <V_i | A0 A| V_j> which has an extra A multiplication.

Now after this fix the `arnoldi` can calculate correctly the eigenpairs of a Hermitian matrix A. In the case where A is non-Hermitian, the deflation in the codes fails to deflate converged subspace because the Ritz vectors are no longer orthogonal.

### DIFF
--- a/itensor/iterativesolvers.h
+++ b/itensor/iterativesolvers.h
@@ -804,9 +804,9 @@ arnoldi(const BigMatrixT& A,
             // "Deflate" previous eigenpairs:
             for(size_t o = 0; o < w; ++o)
                 {
-                //V[j+1] += (-eigs.at(o)*phi[o]*BraKet(phi[o],V[j+1]));
+                //V[j+1] += (-eigs.at(o)*phi[o]*BraKet(phi[o],V[j]));
                 Complex overlap_;
-                gmres_details::dot(phi[o],V[j+1],overlap_);
+                gmres_details::dot(phi[o],V[j],overlap_);
                 V[j+1] += (-eigs.at(o)*phi[o]*overlap_);
                 }
 
--- a/itensor/iterativesolvers.h
+++ b/itensor/iterativesolvers.h
@@ -750,7 +750,7 @@ arnoldi(const BigMatrixT& A,
 
     if(maxsize == 1)
         {
-        if(norm(phi.front()) == 0) randomize(phi.front());
+        if(norm(phi.front()) == 0) phi.front().randomize();
         phi.front() /= norm(phi.front());
         ITensor Aphi(phi.front());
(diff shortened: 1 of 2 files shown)


shencebebetterme:
Right now I know how to reconstruct the correct eigenvectors of A from those of A-A0, but its implementation depends on taking inverse of a matrix (or solving a simple linear equation).  Do we now have some function in ITensor that can deal with the inverse of a general square matrix? If there's one, then I can create another pull request to solve this too.

shencebebetterme:
Other issues of arnoldi:

1. If the matrix A has several degenerate eigenvalues, and the corresponding initial tensors in `phi` vector happen to coincide, then the arnoldi function fails to extract all degenerate eigenvalues. Each degenerate eigenvalue will appear only once, because in this case the number `nh` evaluates to 0 in the deflation process, and the for loop just break regardless of convergence. Maybe we can randomize all the initial tensors in `phi`.

2.  Sometimes the algorithm is not stable and run into local minimum.  This can be shown with the following example
```
A = diag(100, 6, 6, -3) //create such an order-2 ITensor, which is essentially a diagonal matrix
ErrGoal = 1E-10
Max_Iter = 10
Max_Restart = 5
Npass = 1
```
The first 3 eigenvalues can be calculated correctly, but the last one sometimes converges to near -6.0

mtfishman:
Hi @shencebebetterme,

Sorry for the very slow response and thanks for the fix.

I've added some tests (including some placeholder tests for things that are still broken). I'll merge this since it sounds like an improvement, even if it doesn't fully fix `arnoldi`.

Is the orthogonality issue you mention due to the call to `eigen`: https://github.com/ITensor/ITensor/blob/ac2be02c462e5cf2a8bb74fc0c065a263ecafd64/itensor/iterativesolvers.h#L861 outputting non-orthogonal eigenvectors for non-Hermitian matrices?

Cheers,
Matt

mtfishman:
By the way, it looked like the imaginary parts of the eigenvalues were still off in the non-Hermitian cases I tested.

PR #384, opened 2021-02-08
Add error checks for zero blocks in randomITensor QN version

### LINKED ISSUE #383: Should QDense constructor error if zero blocks?
The QDense storage constructor taking an IndexSet and a QN sometimes results in a storage with zero blocks, in the case that the requested QN is not compatible with any settings of the indices. This is ok, but the current behavior is that the constructor just makes no blocks (empty list of blocks) and goes on. As a result, calling `flux` on the resulting ITensor gives `QN()` which is interpreted as a zero QN. Then calling something like `T.generate` inside of a function like `randomITensor` now will generate blocks corresponding to a zero-flux ITensor. So the final QN ends up different from the one requested.

Where should we throw an error to prevent this? Is it reasonable for the QDense constructor to error? It's probably an easy change, but we just need to be sure of the right design.

Example to reproduce:

```
auto sites = SpinHalf(4,{"ConserveQNs=",true});
auto T = randomITensor(QN({"Sz",1}),sites(1),sites(2),sites(3),sites(4));
Print(flux(T) != QN({"Sz",1}));
Print(flux(T) == QN({"Sz",0}));
```

emstoudenmire:


emstoudenmire:
Closes #383

PR #385, opened 2021-03-18
Remove the make warnings on comparing size_t and int values

mingruyang:
.size() returns a unsigned long int, so to detect the overflow it should be compared with the same variable type.


mtfishman:
Thanks Mingru!

PR #386, opened 2021-03-18
If the SVD algorithm gesdd fails, try gesvd instead.

chuffa:
hopefully a simple fix for some larger computations.

### DIFF
--- a/itensor/tensor/algs.cc
+++ b/itensor/tensor/algs.cc
@@ -610,12 +610,12 @@ SVDRef(MatRefc<T> const& M,
             throw std::runtime_error("SVD (ref version), wrong size of D");
 #endif
 
-        auto svdMethod = args.getString("SVDMethod", "gesdd");
+        auto svdMethod = args.getString("SVDMethod", "automatic");
         if(svdMethod=="ITensor")
             {
             SVDRefImpl(M,U,D,V,args);
             }
-        else if(svdMethod == "gesdd" or svdMethod == "gesvd")
+        else if(svdMethod == "automatic" or svdMethod == "gesdd" or svdMethod == "gesvd")
             {
             SVDRefLAPACK(M,U,D,V,args);
             }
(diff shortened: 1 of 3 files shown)


mtfishman:
Great, thanks! Did you check that this fixes issues that you were seeing?

chuffa:
No unfortunately I would have to start a calculation on the cluster, I think sophie might try it, because she has set up everything with the exact parameters that caused the problem.

mtfishman:
Ok, do you want to wait to confirm that this fixes the problems you were seeing before merging? I still think it's useful, but maybe there is something else going on in the case you were seeing as well.

titaschanda:
Hi, I have been using gesdd and gesvd for ITensors for a long time (using a custom svd implementation based on  Armadillo's wrapper of LAPACK, where I have implemented similar fallback). I want to add that gesdd fails pretty often and the parameters for which it happens entirely depends on underlying LAPACK implementation as well as the CPU running it, older CPU without AVX512 fails more often for Intel MKL. However, the above technique of falling back to gesvd always works.

mtfishman:
That's helpful, thanks!

chuffa:
Hi,

As we discussed, I added a default SVDMethod flag called "automatic". An SVD called with this flag first uses gesdd and if that fails it tries gesvd. 
Using any other flag, ITensor tries the specified algorithm, and if it fails it fails.

I also found a problem in my original commit: gesdd destroys the data and if it fails the data needs to be reinitialized.

I tested that all flags work as expected using print statements
I also tested that if gesvd is called after gesdd failed the result is correct (I simply modified the info variable to mimic a failure of gesdd).

For our workflow it would be simplest to put this change into ITensor since I am not runing the calculation that previously failed.

Cheers Daniel

mtfishman:
Thanks Daniel! Looks good, I'll merge. I'll need to add the same feature to the Julia version.

PR #388, opened 2021-03-22
is_pod is going away

tanmoy87544:
is_pod is deprecated in C++20, but we can get ready for that easily: is_trivial && is_standard_layout does the job


emstoudenmire:
Good to know! Thanks for this update to the code.

PR #400, opened 2022-01-06
[C++20] Replace `result_of` with `invoke_result` when using C++ > 17

mtfishman:
For version of C++ greater than 17, replace the use of `result_of` with `invoke_result` since it is removed in C++20.


mtfishman:
@the-hampel @Wentzell

Wentzell:
Are you sure that `__cpp_lib_is_invocable >= 201703` checks the availability of `std::invoke_result` ?

Wentzell:
Ok yes it does, c.f. https://en.cppreference.com/w/cpp/feature_test
Loogs good! Thank you @mtfishman

Issue #303, opened 2019-11-26
Obtaining SiteSet from MPS feature not available in version 3?


HumpyBlumpy:
In ITensor 2, there was a way of obtaining a siteset object of an MPS by calling
`psi.sites()`
for some MPS psi.
This was very useful and I have many functions that make use of this feature. For example, I often use it in such form
`psi.sites().op(op_name,site_num);
`
In trying to upgrade to version 3, calling psi.sites() returns an error and I didnt find another way of obtaining it.
Is it possible to bring this feature back? or is there another way of doing this?

emstoudenmire:
Hi, thanks for the question. So the short answer is this feature just isn’t present in version 3. One reason is that we made it easier to construct MPS without requiring a site set, so then we had a situation where in many cases calling psi.sites() would throw an exception because there was no site set associated with psi. (Things could get especially weird when writing MPS to disk and reading them back, because then code relying on calling psi.sites() wouldn’t work unless one remembered to also separately write the site set and read it back in before reading in the MPS again etc.)

Another reason is that we found it was not too hard to pass the site set, if needed, to the same function as the MPS as just a separate variable. 

So the main fix we recommend is just to pass the site set to any function that needs it as a separate variable. But I do understand that can be a pain sometimes when having a very deep stack of function calls.

Another option is to use the new `siteInds` function added in version 3. Given an MPS `psi`, calling `siteInds(psi)` will return a collection of all of the site indices on it. The only drawback is that this collection is not a site set which use can use to obtain physics operators etc. It’s just an array of Index objects. But it’s definitely very helpful for cases where you need the site indices for writing algorithms.

Ok please let us know if you have any more questions that weren’t answered above.

PR #402, opened 2022-01-13
C++ expect function

JanReimers:
-Uses Matrix1 as the return type. Is there a 2D (matrix) container with dynamically determined base index 
     so row #s line up with site #s?
    -Takes a const MPS& and then copies it so we can move the ortho-centre.Alternately we could work on 
     the original and restore the ortho center before returning.
    -Should test with a complex wave function, easy way to make that?
    -SpinOne seems to be orphaned doesn't use BasicSiteSet<> so it won't work with expect.
    -using std::pair<int,int> for defining the site range.  Range seem to be too advanced for this purpose.
    -Interface is slightly different than the Julia version, site range has to be before the 
     operator strings parameter pack.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ tutorial/finiteT/en.dat
 tutorial/finiteT/metts
 tutorial/finiteT/sus.dat
 TAGS
+
+unittest/test.h5
--- a/itensor/itensor.h
+++ b/itensor/itensor.h
@@ -496,6 +496,16 @@ Cplx
 eltC(ITensor A,
      VarArgs&&... vargs);
 
+//
+//  Templated version that we can call from template code. Simply forwards to elt or eltc based on T.
+//  Real and Complex specializations are in itensor_impl.h
+//
+template <typename T>
+T
+eltT(ITensor A);     
+
+
+
 // Get IndexSet
 IndexSet const& 
 inds(ITensor const& A);
--- a/itensor/itensor_impl.h
+++ b/itensor/itensor_impl.h
@@ -517,6 +517,23 @@ eltC(ITensor A,
     return A.eltC(std::forward<VarArgs>(vargs)...);
     }
 
+//
+//  I was unable to get these to instance with a varargs parameter pack.
+//
+template <> inline
+Real
+eltT(ITensor A)
+    {
+        return A.elt();
+    }   
+
+template <> inline
+Complex
+eltT(ITensor A)
+    {
+        return A.eltC();
+    }   
+    
 //Apply x = f(x) for each element x of T
 //and return the resulting tensor
 template<typename F>
--- a/itensor/mps/mps.cc
+++ b/itensor/mps/mps.cc
@@ -97,7 +97,8 @@ MPS(IndexSet const& sites,
     r_orth_lim_(sites.length()+1),
     atb_(1),
     writedir_("./"),
-    do_write_(false)
+    do_write_(false),
+    pm_sites_(0) //JRTODO Perhaps it is better to make a vector of GenericSites?
     {
     new_tensors(A_,sites,m);
     }
@@ -111,8 +112,9 @@ MPS(InitState const& initState)
     r_orth_lim_(2),
     atb_(1),
     writedir_("./"),
-    do_write_(false)
-    { 
+    do_write_(false),
+    pm_sites_(initState.sites().getPMSites())
+    {
     init_tensors(A_,initState);
     }
 
@@ -125,8 +127,9 @@ MPS(MPS const& other)
     r_orth_lim_(other.r_orth_lim_),
     atb_(other.atb_),
     writedir_(other.writedir_),
-    do_write_(other.do_write_)
-    { 
+    do_write_(other.do_write_),
+    pm_sites_(other.pm_sites_)
+    {
     copyWriteDir();
     }
 
@@ -140,6 +143,7 @@ operator=(MPS const& other)
     atb_ = other.atb_;
     writedir_ = other.writedir_;
     do_write_ = other.do_write_;
+    pm_sites_=other.pm_sites_;
 
     copyWriteDir();
     return *this;
@@ -254,6 +258,7 @@ randomCircuitMPS(SiteSet const& s, int m, Args const& args)
 
     M.leftLim(0);
     M.rightLim(2);
+    M.pm_sites_=s.getPMSites();
     return M;
     }
 
--- a/itensor/mps/mps.cc
+++ b/itensor/mps/mps.cc
@@ -97,8 +97,7 @@ MPS(IndexSet const& sites,
     r_orth_lim_(sites.length()+1),
     atb_(1),
     writedir_("./"),
-    do_write_(false),
-    pm_sites_(0) //JRTODO Perhaps it is better to make a vector of GenericSites?
+    do_write_(false)
     {
     new_tensors(A_,sites,m);
     }
@@ -112,9 +111,8 @@ MPS(InitState const& initState)
     r_orth_lim_(2),
     atb_(1),
     writedir_("./"),
-    do_write_(false),
-    pm_sites_(initState.sites().getPMSites())
-    {
+    do_write_(false)
+    { 
     init_tensors(A_,initState);
     }
 
@@ -127,9 +125,8 @@ MPS(MPS const& other)
     r_orth_lim_(other.r_orth_lim_),
     atb_(other.atb_),
     writedir_(other.writedir_),
-    do_write_(other.do_write_),
-    pm_sites_(other.pm_sites_)
-    {
+    do_write_(other.do_write_)
+    { 
     copyWriteDir();
     }
 
@@ -143,7 +140,6 @@ operator=(MPS const& other)
     atb_ = other.atb_;
     writedir_ = other.writedir_;
     do_write_ = other.do_write_;
-    pm_sites_=other.pm_sites_;
 
     copyWriteDir();
     return *this;
@@ -258,7 +254,6 @@ randomCircuitMPS(SiteSet const& s, int m, Args const& args)
 
     M.leftLim(0);
     M.rightLim(2);
-    M.pm_sites_=s.getPMSites();
     return M;
     }
 
--- a/itensor/mps/mps.h
+++ b/itensor/mps/mps.h
@@ -871,8 +871,8 @@ sum(std::vector<MPSType> const& terms,
 //  Template implementation of expect function for Real and Complex types and fixed site list.
 //
 template <class T> std::vector<std::vector<T>> 
-expectT(const SiteSet& sites,
-        const MPS& _psi, 
+expectT(const MPS& _psi, 
+        const SiteSet& sites,
         const std::vector<string>& vops,
         std::vector<int> site_list 
         )
@@ -910,8 +910,8 @@ expectT(const SiteSet& sites,
 //  Convert range to explicit list in one place
 //
 template <class T> std::vector<std::vector<T>> 
-expectT(const SiteSet& sites,
-        const MPS& _psi, 
+expectT(const MPS& _psi, 
+        const SiteSet& sites,
         const std::vector<string>& vops,
         detail::RangeHelper<int> site_range=range1(0) //fake default because we don't have access to sites.length in the function signature.
         )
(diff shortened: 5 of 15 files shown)


mtfishman:
Hi @JanReimers, thanks for looking into this.

> -Uses Matrix1 as the return type. Is there a 2D (matrix) container with dynamically determined base index
> so row #s line up with site #s?

An alternative interface would be the following. Split the function into two versions:

1. One version takes an `std::vector<string>`, and outputs a vector of results, i.e. `std::vector<std::vector<Real>>`.
2. A second version just takes a single `string` input and outputs a single `std::vector<Real>`, and is implemented as a simple wrapper around version 1.

In fact, I think this is how the Julia version of `expect` should be implemented as well.

> -Takes a const MPS& and then copies it so we can move the ortho-centre.Alternately we could work on
> the original and restore the ortho center before returning.

The current design you describe is generally how we handle these things. Copying the MPS does not copy the tensor data, so it is cheap.

> -Should test with a complex wave function, easy way to make that?

`randomMPS` takes a `Complex` argument:

https://github.com/ITensor/ITensor/blob/684bb4207da536dce59f1f2589b60ec23d1d61cb/unittest/mpo_test.cc#L142

How are you handling the complex version (i.e. a version where the actual output is complex, like getting the expectation value of a non-Hermitian operator)? I think in general we handle this kind of thing by having a version `expect` that outputs a real value and `expectC` that outputs a complex value, where they are implemented as wrappers around a single internal implementation.

> -SpinOne seems to be orphaned doesn't use BasicSiteSet<> so it won't work with expect.

I admit I haven't looked at this code in a while, could you expand on this issue? Could we modify the definition of `SpinOne` to fix this?

> -using std::pair<int,int> for defining the site range. Range seem to be too advanced for this purpose.

Perhaps we could make it more general and accept a `std::vector<int>`, or make it generically anything that is iterable?

> -Interface is slightly different than the Julia version, site range has to be before the
> operator strings parameter pack.

That's ok, but could be fixed by my design proposal for the interface above.

I have two larger concerns about the design:

1. I remember we previously had the `SiteSet` stored inside the MPS and have since removed it. The main reason for removing it was that an MPS can be modified in ways such that the original `SiteSet` it was constructed from is no longer valid (like setting a tensor element with a new tensor that has a different site type). Additionally, an MPS can just be created as empty, and then have the elements set with tensors (which is a related issue). It seemed like a can of worms to try to track every operation where the MPS might be modified or operated on and update the `SiteSet` appropriately. I would actually prefer to have the `SiteSet` passed to the `expect` function over introducing a difficult-to-maintain feature of having the `SiteSet` in the MPS class. See also #303.
2. It seems like you've made some deeper modifications to the `SiteSet` design. I have not thought deeply about that system in a while, that system was designed and maintained primarily by @emstoudenmire. Perhaps he can comment about modifications to that part of the code. Could you explain why that code had to be modified?

JanReimers:
Hey thanks for the quick analysis and reply.  See below:

On Fri, Jan 14, 2022 at 10:45 AM Matt Fishman ***@***.***>
wrote:

> Hi @JanReimers <https://github.com/JanReimers>, thanks for looking into
> this.
>
> -Uses Matrix1 as the return type. Is there a 2D (matrix) container with
> dynamically determined base index
> so row #s line up with site #s?
>
> An alternative interface would be the following. Split the function into
> two versions:
>
>    1. One version takes an std::vector<string>, and outputs a vector of
>    results, i.e. std::vector<std::vector<Real>>.
>    2. A second version just takes a single string input and outputs a
>    single std::vector<Real>, and is implemented as a simple wrapper
>    around version 1.
>
> 
Yeah we can do that.  If we want anything other than 0 based for the
output then we have to pad with zeros which might surprise users.

>
>
> In fact, I think this is how the Julia version of expect should be
> implemented as well.
>
> -Takes a const MPS& and then copies it so we can move the
> ortho-centre.Alternately we could work on
> the original and restore the ortho center before returning.
>
> The current design you describe is generally how we handle these things.
> Copying the MPS does not copy the tensor data, so it is cheap.
>
OK good.

> -Should test with a complex wave function, easy way to make that?
>
> randomMPS takes a Complex argument:
>
>
> https://github.com/ITensor/ITensor/blob/684bb4207da536dce59f1f2589b60ec23d1d61cb/unittest/mpo_test.cc#L142
>
> How are you handling the complex version (i.e. a version where the actual
> output is complex, like getting the expectation value of a non-Hermitian
> operator)? I think in general we handle this kind of thing by having a
> version expect that outputs a real value and expectC that outputs a
> complex value, where they are implemented as wrappers around a single
> internal implementation.
>
Yes we could make a second expectC function. We could use templates behind
the scenes to avoid too much code duplication.

> -SpinOne seems to be orphaned doesn't use BasicSiteSet<> so it won't work
> with expect.
>
> I admit I haven't looked at this code in a while, could you expand on this
> issue? Could we modify the definition of SpinOne to fix this?
>
I just noticed SpinTwo is the same as SpinOne. All the other sitesets use a
template:
using SpinHalf = BasicSiteSet<SpinHalfSite>;
Where as SpinOne and SpinTwo site sets are hand coded.  I just tried
switching them over to the template mechanism and all the unit tests pass.
So maybe it is trivial to get everything consistent, but I will have to
look more closely to be sure.

> -using std::pair<int,int> for defining the site range. Range seem to be
> too advanced for this purpose.
>
> Perhaps we could make it more general and accept a std::vector<int>, or
> make it generically anything that is iterable?
>
The pair was intended to be a first and last site (a range) as opposed to a
list of sites like {1,3,7,11} to include.  If we want use syntax like this
expect(psi,{2,6},"Sz");
then we have to pick one or the other.  If we try and support both ideas I
think users will get confused.

> -Interface is slightly different than the Julia version, site range has to
> be before the
> operator strings parameter pack.
>
> That's ok, but could be fixed by my design proposal for the interface
> above.
>
Yes, as soon as we get away from varargs then our options for the function
signature become less constrained.

> I have two larger concerns about the design:
>
Excellent ... I was concerned about this as I was coding, but I decided
that instead of bombarding you guys with email, I would propose something
concrete for discussion.

>
>    1. I remember we previously had the SiteSet stored inside the MPS and
>    have since removed it. The main reason for removing it was that an MPS can
>    be modified in ways such that the original SiteSet it was constructed
>    from is no longer valid (like setting a tensor element with a new tensor
>    that has a different site type). Additionally, an MPS can just be created
>    as empty, and then have the elements set with tensors (which is a related
>    issue). It seemed like a can of worms to try to track every operation where
>    the MPS might be modified or operated on and update the SiteSet
>    appropriately. I would actually prefer to have the SiteSet passed to
>    the expect function over introducing a difficult-to-maintain feature
>    of having the SiteSet in the MPS class. See also #303
>    <https://github.com/ITensor/ITensor/issues/303>.
>
> 
OK this is very good to know.  THe reason this came up is that there is no
C++ version op("Sz",Index(2, "Site,S=1/2")).  To get the op we need to know
the hilbert space which is known in the SiteSet but not in psi which only
knows the IndexSet and d.  Miles hinted that the julia version somehow uses
reflection (introspection) to get the job done.  In my way of thinking, if
the MPS objects don't know the hilbert space (all they know is d) then they
really are a Matrix Product and not a Matrix Product State.  IN other
words, just coefficients with no kets, i.e no context.
But getting back out of the philosophy rabbit hole: I agree passing the
SiteSet for corresponding to psi in the expect call is the short term
solution.
I think if there are situations where we should not have hilbert space info
inside MPS, then it would be very beneficial to make some unit tests that
demonstrate the problem in concrete terms.  That will quickly steer myself
and others away from breaking anything in this regard. If you agree I will
try and add some tests that demonstrate the use cases you described above.

>
>    1. It seems like you've made some deeper modifications to the SiteSet
>    design. I have not thought deeply about that system in a while, that system
>    was designed and maintained primarily by @emstoudenmire
>    <https://github.com/emstoudenmire>. Perhaps he can comment about
>    modifications to that part of the code. Could you explain why that code had
>    to be modified?
>
> 
Miles indicated that it would be good to simplify the SiteSet code.  I
think we can get rid of SiteHolder and SiteStore. The key is to have all
concrete site types virtually derived from the abstract base SiteBase.  In
order to get operators I need to make a virtual call to
SiteBase->op("Sz").  Now that I have studied this more I think I can do
what we want with less intrusion into SiteSet.  The changes I made were
intended to be the first step in refactoring/simplifying SiteSet.

mtfishman:
> Yeah we can do that. If we want anything other than 0 based for the output then we have to pad with zeros which might surprise users.

I'm not sure if we need 1 based indexing in this case. There is definitely an awkward clash between 0-based indexing used by C++ and 1-based indexing used by ITensor. Perhaps the way to think about this case is that really people will want to iterate through the resulting data in some way (make a plot, find an average, etc.) in which case the indexing offset is not important, so we may as well stick with the convention of whatever data structure we return (in this case, use 0-based for an `std::vector`).

> Yes we could make a second expectC function. We could use templates behind the scenes to avoid too much code duplication.

Yeah, that's a good strategy.

> I just noticed SpinTwo is the same as SpinOne. All the other sitesets use a template: using SpinHalf = BasicSiteSet<SpinHalfSite>; Where as SpinOne and SpinTwo site sets are hand coded. I just tried switching them over to the template mechanism and all the unit tests pass. So maybe it is trivial to get everything consistent, but I will have to look more closely to be sure.

Ok, perhaps it was just a mistake that they weren't defined in the same way as the other site types, so if it doesn't break anything we should just make them consistent.

> The pair was intended to be a first and last site (a range) as opposed to a list of sites like {1,3,7,11} to include. If we want use syntax like this expect(psi,{2,6},"Sz"); then we have to pick one or the other. If we try and support both ideas I think users will get confused.

Right, I think we should make it templated so it accepts a range of sites like `{2, 3, 4, 5, 6}`, or a range of the form `range(2, 7)` (really anything you can iterate over in a `for` loop).

> OK this is very good to know. THe reason this came up is that there is no C++ version op("Sz",Index(2, "Site,S=1/2")). To get the op we need to know the hilbert space which is known in the SiteSet but not in psi which only knows the IndexSet and d. Miles hinted that the julia version somehow uses reflection (introspection) to get the job done. In my way of thinking, if the MPS objects don't know the hilbert space (all they know is d) then they really are a Matrix Product and not a Matrix Product State. IN other words, just coefficients with no kets, i.e no context. But getting back out of the philosophy rabbit hole: I agree passing the SiteSet for corresponding to psi in the expect call is the short term solution. I think if there are situations where we should not have hilbert space info inside MPS, then it would be very beneficial to make some unit tests that demonstrate the problem in concrete terms. That will quickly steer myself and others away from breaking anything in this regard. If you agree I will try and add some tests that demonstrate the use cases you described above.

Yes, in Julia it was technically feasible to encode the type of the site in the tag. In my opinion that was a huge innovation for the ITensor project which was enabled by the dynamic types and multiple dispatch system in Julia, and required a lot of design discussions by Miles and me!

As an argument against your philosophical discussion about the design of the MPS type, I would argue that Hilbert spaces are defined through their dimension (for the case of quantum number conservation, you have a [graded vector space](https://en.wikipedia.org/wiki/Graded_vector_space) which are then defined by the sector dimensions and symmetry labels). So whether or not an MPS has the type of the sites (S=1/2, Electron, etc.) saved inside it isn't core to an MPS being a state or not, those are more just conventions which define which types of operators you want to define on the sites of the MPS (so more like defining the local operator basis on the space of your MPS).

But as for the practical issues with the design, it would make it challenging to write code like the following:
```julia
  auto psi = MPS(10);
  for(auto i : range(length(psi)))
    { 
    psi.ref(i) = ITensor(Index(2));
    }
```
More generally, there are cases where you have an MPS where the site indices don't have an interpretation as an existing site type, for example when you contract the inner product of a PEPS (or a 2D classical partition function), a common contraction strategy is to contract from the boundary using an MPS, but the sites of the MPS are the (doubled) link dimensions of the PEPS, which have no direct interpretation in terms of a site type.

Additionally, there are cases where you dynamically change the space of the Hilbert space, like when you project an Electron site type into a SpinHalf site type with a Gutzwiller projection, or if you are doing gate evolution of an MPS with mixed types of sites (such as a combination of SpinHalf and Fermion) with long range interactions, where the swap gates used to implement the long range gates swap the ordering of the sites of the MPS dynamically.

We can make a FAQ about this issue of putting the SiteType inside the MPS, but in the end for the sake of minimizing developer burden and the above examples where that design does not work well, I would say definitively that we are not going to do it at this point, and passing around the SiteType as needed is easy enough for users.

> Miles indicated that it would be good to simplify the SiteSet code. I think we can get rid of SiteHolder and SiteStore. The key is to have all concrete site types virtually derived from the abstract base SiteBase. In order to get operators I need to make a virtual call to SiteBase->op("Sz"). Now that I have studied this more I think I can do what we want with less intrusion into SiteSet. The changes I made were intended to be the first step in refactoring/simplifying SiteSet.

Why don't we leave those changes to a future PR? It definitely sounds like a worthy goal to simplify the SiteSet code.

mtfishman:
The latest changes look good, thanks!

A very small change I would make would be to order the function signature as `expect (const MPS& _psi, const SiteSet& sites, ...)` instead of `expect (const SiteSet& sites, const MPS& _psi, ...)`. I think that ordering makes more sense, since the `SiteSet` is conceptually more closely associated with the operators. This is just a personal preference, open to arguments the other way.

Also, are you planning on adding a wrapper which accepts only a single `string op` and outputs a single `vector` of measurements?

JanReimers:
Hi Matt, I just updated PR402 with the changes you suggested.

Regards
Jan

mtfishman:
Looks great, thanks Jan! Anything else left to do? Otherwise I will merge.

JanReimers:
Hi Matt, nothing more that I know of.  So yeah go ahead and merge.

PR #405, opened 2022-01-29
correlationMatrix v1.0

JanReimers:
-Take isHermitian=fals flag when user expects a non-Hermitian corr. matrix
-Throw an exception if the user supplies an unsupported mixed fermionic*Bosonic operator like A*Cdag

### DIFF
--- a/itensor/mps/mps.cc
+++ b/itensor/mps/mps.cc
@@ -15,6 +15,7 @@
 //
 #include "itensor/mps/mps.h"
 #include "itensor/mps/mpo.h"
+#include "itensor/mps/autompo.h" //need this for SiteTerm
 #include "itensor/mps/localop.h"
 #include "itensor/util/print_macro.h"
 #include "itensor/util/str.h"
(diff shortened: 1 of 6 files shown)


itensor/mps/mps.cc
```diff
@@ -1537,6 +1538,167 @@ inner(MPS const& psi, MPS const& phi) //Re[<psi|phi>]
...
+correlationMatrixT(const MPS& _psi,
+                   const SiteSet& sites,
+                   const string& _op1,
+                   const string& _op2,
+                   detail::RangeHelper<int> site_range,
+                   Args const& args
+                  )
+{
+    assert(checkConsistent(_psi,sites));
+    bool isHermitian= args.getBool("isHermitian",true);
```
mtfishman: Should this default to `true` or `false`?

I would vote for `false` to be on the safe side and avoid surprising behavior if the operators don't commute.

@emstoudenmire what do you think?

JanReimers:
Yes it should be false, you even said so in your email ... and I missed it.  It is  a little tricky for users because things like <S+S-> just happen to be symmetric.  If we default to false we could switch it to true if op1==op2 *and* the user didn't intentionally specify false.

mtfishman:
That's a great suggestion, we could have a special check in cases where we know the correlation matrix is symmetric (like when the operators are equal, and we could hard-code a few special cases like `Cdag C` and `S+ S-`). That should make it so it is efficient in most cases people care about, but also safe.

A potential issue with hard-coding some cases like `Cdag C` and `S+ S-` is that someone could define their own version of those operators for a new site type where those operators do not lead to symmetric correlation matrices. In Julia we could handle this with a trait-like function:
```julia
# General case where the operators are the same
is_correlation_matrix_hermitian(::SiteType, ::ON, ::ON) where {ON<:OpName} = true

# Special cases
is_correlation_matrix_hermitian(::SiteType"Fermion", ::OpName"Cdag", ::OpName"C") = true
is_correlation_matrix_hermitian(::SiteType"S=1/2", ::OpName"S+", ::OpName"S-") = true
```
Not sure how easy it would be to have a similar function overloading style in C++ to "register" a characteristic of a correlation matrix made from certain operators with a certain site type.

JanReimers:
I am guessing <S+S-> is symmetric because of time reversal symmetry, creating boson at site _i_ and destroying it at _j_ is the same creating an anti boson at _j_ and destroying it at _i_.  I don't think it is the job of the correlation Matrix function to interpret these strings in any deep way for the reasons you state, extensibility of site physics.  However it could ask a question of the site like site.isTimeSymmetric("S+","S-") but I think that should be out of scope for this commit.

mtfishman:
Sounds good, agreed that analyzing the operators for symmetry is a deeper topic and doesn't need to be handled in this PR. So in this PR, we could just default to `false` and then set it to `true` if the operators are the same.

JanReimers:
Will do.  Thanks

mtfishman:
Great, thanks @JanReimers! Over all it looks good to me.

Let's see if @emstoudenmire has any comments about the `isHermitian` design or anything else in the PR, otherwise I'm good to merge this.

JanReimers:
Sure.  There are a couple of nit picky things that might be improvable:
1)
        auto c=Li * sites.op(onsiteOp,i) *
dag(noPrime(prime(psi(i)),prime(rind)) )/ norm2_psi;
in Julua you used something like
        c=Li * sites.op(onsiteOp,i) * prime(dag(psi[i]), *not(rind)*))/
norm2_psi;
which is much more readable: prime everything BUT rind.  Is there a nice
clean way to do that on the C++ side?

2) Is it better to just normalize the copy of psi instead of dividing by
norm2_psi everywhere?
3) Error policies: To abort or to throw?  I chose throw simply because then
I could test it in the catch2 framework which doesn't have any obvious
mechanism for expecting abort.  But that is probably not the best reason!
4) Should we throw an exception instead of assert (abort) if the MPS and
SiteSet are inconsistent?

Regards
Jan

mtfishman:
> Sure. There are a couple of nit picky things that might be improvable: 1) auto c=Li * sites.op(onsiteOp,i) * dag(noPrime(prime(psi(i)),prime(rind)) )/ norm2_psi; in Julua you used something like c=Li * sites.op(onsiteOp,i) * prime(dag(psi[i]), *not(rind)*))/ norm2_psi; which is much more readable: prime everything BUT rind. Is there a nice clean way to do that on the C++ side? 

We don't have the `not` functionality available in the C++ version. The priming/tagging functionality was completely rewritten in the Julia to make it much easier to implement that kind of functionality (the generic programming and multiple dispatch features in Julia helped a lot with that too). I think that code is fine how it is, you could just separate that part into two lines like:
```C++
auto psi_i_dag = dag(noPrime(prime(psi(i)), prime(rind)));
auto c = Li * sites.op(onsiteOp,i) * psi_i_dag / norm2_psi;
```
to make it a bit more readable. Something else I've done in cases like that is used logic based on index set functions like `uniqueInds` and `commonInds`. For example, maybe something like:
```C++
auto linds = uniqueInds(psi(i), psi(i+1));
auto psi_i_dag = dag(prime(psi(i), linds)));
```

> 2) Is it better to just normalize the copy of psi instead of dividing by norm2_psi everywhere?

I don't think it really matters either way, maybe it is slightly cheaper to divide by the norm within the function instead of normalizing psi at the beginning since it reduces to scalar arithmetic but either way it is a subleading cost.

> 3) Error policies: To abort or to throw? I chose throw simply because then I could test it in the catch2 framework which doesn't have any obvious mechanism for expecting abort. But that is probably not the best reason!

I have no strong opinion about this, the only time I've actually used the specific type error is in tests so I guess whatever is best for testing is fine.

> 4) Should we throw an exception instead of assert (abort) if the MPS and SiteSet are inconsistent? 

Again, I don't have much of an opinion on it, I guess whichever makes it easiest to test that functionality.

itensor/mps/mps.cc
emstoudenmire: Thanks Matt. I was thinking about this issue in the context of the Julia code too, since you raised it there too. 

What do you both think of this idea: calling the two operators A and B, if A is the Hermitian conjugate of B, then we can detect `isHermitian=true` automatically and do the more efficient algorithm, otherwise do the less efficient one. We could just put in a check like `norm(A - dag(swapPrime(B,0,1))) < 1E-10` or similar to check whether A is the conjugate of B. Since A and B are small (low-dimensional) operators, a numerical check like this ought to be reliable I think. If for some reason the check fails, all that will happen is a slightly slower algorithm will run but the output would still be correct.

emstoudenmire: We could still let the user specifiy `isHermitian` to override this check when they are sure about it, and to ensure a certain level of performance, but then the default should be `false`.

emstoudenmire:
I just caught that you both discussed automatic checks for `isHermitian` some more already. I think a sufficient condition for Hermitian-ness is that the operators are conjugates of each other. There may other cases where they are not but somehow the state itself has special properties (trivial case is if the state is 0!) but we shouldn't check for those kinds of cases. 

It does make a good argument for allowing the user to optionally specify `isHermitian` if they are sure that will be true.

itensor/mps/mps.cc
mtfishman: @emstoudenmire thanks for the comment, I didn't consider the idea of forming the operator and checking the Hermiticity of the tensor. That is a more general and automated solution than the one I brought up.

I think it is fine to check the Hermiticity up to some tolerance. I guess more specifically the operators only need to be Hermitian up to a scale factor, so we could check the Hermiticity of the normed operators:
```C++
norm(A / norm(A) - dag(swapPrime(B / norm(B), 0, 1))) < 1E-10
```

mtfishman:
Thanks @JanReimers, looks great! Is there a reason you've removed `auto` in favor of explicit types? In general we have preferred the style of using `auto`.

JanReimers:
Ok I put in a Hermitian test with round off tolerance as per Matts suggestion.  With this test users should rarely if ever need to use the isHermitian override flag.

JanReimers:
I left in auto for all the ints as my patience for fussing with long and
unsigned etc is precisely zero.  BUT for ITensor classes it helps me to see
exactly what is going on, i.e. is it returning an Index or an IndexSet
etc.  As I gain experience it will become totally obvious what all those
auto types must be.  I can put them back to auto if you prefer that.

Regards
Jan

mtfishman:
Indeed, for the case of ints things are quite annoying.

For the case of ITensor objects, we've tried to keep the naming very consistent (functions with `*Index` return an `Index`, functions with `*Inds` return an `IndexSet`, etc.).

I don't have a strong opinion either way, mostly I was curious. In Julia, it is kind of like there is an `auto` by default everywhere, so I've gotten used to the style of just leaving off types everywhere. The extra flexibility adds some burden to think through the types yourself, but well written generic code and good library design (good function names, writing type-stable code) makes it so that users shouldn't have to think too hard about the types and things should "just work".

Anyway, this looks good from my end, ready to merge?

JanReimers:
Yes I found auto to be a game changer when using and creating expression
template code on C++.  Yes, good to merge from my end.


Regards
Jan

mtfishman:
Thanks Jan, this will be very useful to have! If you are up for it, we would like to have the same Hermitian/non-Hermitian logic in the Julia version as well (since you've already written it in the C++ version it should be pretty easy to port the same logic to the Julia version).

JanReimers:
I was just thinking the same thing ... sure.  There will be a delay,
because I have to fork the julia side and get git to do what I want!

Regards
Jan

mtfishman:
Thanks, no rush!

mtfishman:
By the way, I've added a more in-depth guide to developing ITensors.jl here: https://itensor.github.io/ITensors.jl/stable/AdvancedUsageGuide.html#Developing-ITensors.jl

PR #406, opened 2022-01-30
Allow densityMatrixApplyMPOImpl to write environment tensors to disk …

mingruyang:
…and reduce memory usage. This would be very helpful for the global subspace expansion of TDVP.

### DIFF
--- a/itensor/mps/mpoalgs.cc
+++ b/itensor/mps/mpoalgs.cc
@@ -193,7 +193,7 @@ MPS
 densityMatrixApplyMPOImpl(MPO const& K,
                           MPS const& psi,
                           Args args)
-    {
+    {       
     if( args.defined("Maxm") )
       {
       if( args.defined("MaxDim") )
@@ -214,6 +214,9 @@ densityMatrixApplyMPOImpl(MPO const& K,
     dargs.add("RespectDegenerate",args.getBool("RespectDegenerate",true));
     auto verbose = args.getBool("Verbose",false);
     auto normalize = args.getBool("Normalize",false);
+    auto dowrite = args.defined("WriteDim") && (maxLinkDim(psi) >= args.getInt("WriteDim"));
+    std::string writedir_ = "./";
+    if(dowrite) writedir_ = mkTempDir("E","./");
 
     auto N = length(psi);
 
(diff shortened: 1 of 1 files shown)


itensor/mps/mpoalgs.cc
```diff
@@ -225,33 +228,39 @@ densityMatrixApplyMPOImpl(MPO const& K,
...
-    for(auto j : range1(N-1)) 
-        Kc.ref(j).prime(-rand_plev,uniqueSiteIndex(Kc,psic,j));
+    //for(auto j : range1(N-1)) 
+    //    Kc.ref(j).prime(-rand_plev,uniqueSiteIndex(Kc,psic,j));
 
     //Build environment tensors from the left
     if(verbose) print("Building environment tensors...");
     auto E = std::vector<ITensor>(N+1);
-    E[1] = psi(1)*K(1)*Kc(1)*psic(1);
+    E[1] = psi(1)*K(1)*dag(K(1)).prime(rand_plev).prime(-rand_plev,uniqueSiteIndex(K,psi,1).prime(rand_plev))*dag(psi(1)).prime(rand_plev);
```
emstoudenmire: I find this version harder to read and understand than the previous version. I think you're right, though, that this version uses less memory as it runs because of the "on the fly" calls to dag here, rather than in advance. However, the advance calls to priming functions would not require more memory since priming an ITensor does not copy its data.

So I'd propose the following change: restoring the previous version using psic and Kc but only making them different by a prime level and not "dag" ing them ahead of time. Then in the loop, putting the calls to dag around the invididual elemenst of Kc and psic.

emstoudenmire (changes requested):
Thanks for this nice improvement. Let's discuss the one comment and then I'd be happy to pull this in.

itensor/mps/mpoalgs.cc
mingruyang: Thanks for clarifying!

I'm not super familiar with how the data is stored in MPS/MPO and still have three questions:
1. I think the copying of data happens at "auto psic = psi;auto Kc = K;", and the ".dag()" and ".prime()" won't copy the data, is that correct?
2. The difference of dag(psi) or prime(psi) and psi.dag() or psi.prime() is that psi.dag() or psi.prime() does not copy the data while dag(psi) or prime(psi) copies the data, is that correct?
3. Your suggestion is doing "auto psic = prime(psi);auto Kc = prime(K)" ahead of the loop. If 2. is true, then will this still copy the data of the whole MPS and MPO at a time?

emstoudenmire: Good questions. The difference about copying here actually not about the 'free' versus class-method versions (so not between `dag(..)` and `.dag(..)`) but between `dag` and `prime`. Because the indices of an ITensor are stored separately from its data, priming does not copy the data. But calling `dag` will always copy the data if it is complex-valued.

So I think the best version of the code would only do priming ahead of time, with lines of code like `Kc.prime(rand_plev)` and then inside the loop still leaving calls like `dag(K(j))` as you have now, which should reduce memory usage because those dag's are done on the fly and won't be copying a lot of data all at once at the beginning.

mingruyang: Does "auto psic = psi;auto Kc = K;" copy the data?

emstoudenmire: Oh yes, meant to answer that question also. No, it does not copy the data. Copying an MPS or MPO copies its ITensors, but copying an ITensor only copies the smart pointer (shared_ptr) inside which references the data. So copying an ITensor is like copying a reference variable in Python or Julia. Only when you modify an ITensor (like calling dag on a complex-valued ITensor) will the data finally be automatically copied underneath.

mingruyang: Thanks a lot for the explanation!

mingruyang: Now the prime is done before the loop. Actually I have another question: will E[j-1] = ITensor() destruct the tensor previously in E[j-1] and free the memory? How?

emstoudenmire: Great. Yes it will destruct the tensor in this case, though it depends on how E[j-1] was created or used. The way it works is that E[j-1], which is an ITensor object, holds a shared_ptr inside to its data. If E[j-1] is never copied then the reference counter inside the shared_ptr will have the value 1 to say that only one ITensor is sharing that data. When E[j-1] gets overwritten, the reference counter goes down by 1, so to 0, and then the shared_ptr deletes the data that it refers to. But if an earlier line in the code copied E[j-1] into a separate ITensor, then the reference counter would be greater than 1 and overwriting E[j-1] would not free the data.

I think in the code pattern used here, E[j-1] is the only ITensor sharing its data (reference count of 1) so overwriting it will free its data. It's actually possible to check the reference count of an ITensor as a "correctness check" to make sure that the data does get freed right away. We could make a function that does that i.e. confirm that the data gets freed or else throw an exception.

emstoudenmire (approved):

Issue #392, opened 2021-04-24
Two issues in installing ITensor c++ on Windows


YuhanLiuSYSU:
I am using windows 10 with Cygwin for installation and here are two issues I encounter, along with my (might not be the most elegant) solutions:

### Issue 1
The `Makefile` in itensor folder does not write on `this_dir.mk`. To resolve it, I replace `@echo "THIS_DIR=$(PWD)"` by
```
@echo "THIS_DIR=$(shell pwd)"
```
### Issue 2
When compling `algs.cc` , I got the error:
```
In file included from /usr/lib/gcc/x86_64-pc-cygwin/10/include/c++/cctype:42,
                 from /usr/lib/gcc/x86_64-pc-cygwin/10/include/c++/bits/localefwd.h:42,
                 from /usr/lib/gcc/x86_64-pc-cygwin/10/include/c++/string:43,
                 from /usr/lib/gcc/x86_64-pc-cygwin/10/include/c++/stdexcept:39,
                 from tensor/algs.cc:17:
tensor/algs.cc: In function 'int itensor::detail::hermitianDiag(int, itensor::Real*, itensor::Real*)':
tensor/algs.cc:37:20: error: expected unqualified-id before numeric constant
   37 |         LAPACK_INT _N = N;
      |                    ^~
```
My solution is simply replacing `_N` and `_Rrows` by names that don't contain underscore. I don't fully understand the reason behind and I still wonder what cause this problem.

I am a very beginner to makefile so these might be really trivial. But maybe it can help other beginners :)

lgds:
Hi. I ran into the same issue as YuhanLiuSYSU when I updated ITensor on Cygwin.  I did a quick Google search and got here.
I checked that the previous version of algs.cc did not have underscores in front of "N" or "M" and "Rrows" etc. so I just commented out lines like
//LAPACK_INT _M = M, _N = N, _Rrows = Rrows;
and replaced "_N" with "N", "_M" with "M" etc. everywhere. It compiles ok now. 
I hope I did not break anything!...

mtfishman:
Could one of you make a pull request with the minimal changes needed to get it working on Cygwin? If the tests pass we can merge it.

lgds:
Would it help if I placed the file here?
[algs_compilesCygwin.cc.txt](https://github.com/ITensor/ITensor/files/10909981/algs_compilesCygwin.cc.txt)

mtfishman:
[...]

PR #417, opened 2023-03-31
Renamed local variables starting with an underscore followed by a capital letter

haggaila:
The C++ standard reserves identifiers starting with an underscore and then a capital letter (or two underscores) to the compiler implementations. Indeed on Windows with Cygwin64 the macro `_N` (and possibly others) is defined (externally to ITensor), such that occurrences of this string in the code get replaced and the ITensor `algs.cc` preprocessed file does not pass compilation.
This PR renames `_N`, `_M`, and `_Rrows`.

### DIFF
--- a/itensor/tensor/algs.cc
+++ b/itensor/tensor/algs.cc
@@ -34,49 +34,49 @@ namespace detail {
     int
     hermitianDiag(int N, Real *Udata, Real *ddata)
         {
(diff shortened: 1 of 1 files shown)


haggaila:
Solves item (2) in issue #392 .

mtfishman:
Thanks!

